PR #4, opened 2026-06-28
fix(server): harden session lifecycle, send lock, and Cease handling

bashrusakh:
## Summary

P1 fixes from `FIXPLAN.md`: race on session replace, parallel writes to `_stream`, missing Hold Timer, NOTIFICATION on graceful termination, `_state` without memory barriers.

## Why

These are correctness bugs that can cause session resets, route leaks, or protocol violations:

- 1.1 — `_sessions[peerAddress] = session` clobbered the new session when the old session's finally-block ran.
- 1.2 — `NetworkStream` is not thread-safe; `_sendLock` was only held in `RefreshRoutesAsync`, so keepalive / initial routes / NOTIFICATION could interleave bytes.
- 1.3 — Hold Timer was never observed; the dead `BgpTimers.cs` POCO was unused.
- 1.4 — Outer `catch (Exception)` did not emit NOTIFICATION/Cease (RFC 4271 §8.1 violation).
- 1.5 — `IsEstablished` read `_state` from another thread without a memory barrier.

## What changed

- `BGPLite.Server/BgpServer.cs` — `TryAdd` + `ReferenceEquals` guard + NOTIFICATION/Cease to older session on replacement conflict.
- `BGPLite.Server/BgpSession.cs` — single `SendMessageAsync` synchronized entry point; split `_advertisedPrefixesLock` from `_sendLock`; `volatile BgpFsmState _state`; `_ceaseSentOnTeardown` latch (int + `Interlocked`) shared between `RunAsync` catch paths and the `finally` block.
- `BGPLite.Server/BgpTimers.cs` — **deleted** (was a dead 10-line POCO; `HoldTimerLoopAsync` is the real implementation and lives next to the FSM).
- `BGPLite.Tests/BgpSessionShutdownTests.cs` — new regression test `HoldTimer_Expiry_Emits_Notification_HoldTimerExpired` covering P1.3 + P1.4.

## Validation

- `dotnet build BGPLite.Tests` — 0 errors (1 pre-existing CS8602 in `BgpMessageTests.cs`, unchanged).
- `dotnet test BGPLite.Tests` — **107/107 passed** (97 prior + 1 new P1 + 9 from `origin/main` `RipeStat*`).
- `dotnet format whitespace/style/analyzers --verify-no-changes` — clean on changed files.

## Risk / regressions

- Low. `_ceaseSentOnTeardown` switched from `bool` to `int` + `Interlocked.Exchange`; same semantics, but now correct across the external-then-internal-call boundary (BgpServer → NotifyCeaseAsync on shutdown / session replace).
- `BgpTimers.cs` deletion: confirmed no external references via grep before removal.
- Hold-timer detection granularity is `keepAliveInterval` (Math.Max(holdTime/3, 1)) so expiry is observed within `holdTime + keepalive` seconds, acceptable per RFC 4271.

## Out of scope

P2 (protocol validation: path attribute order, `PrefixCodec` length guard, OPEN payload length, AS_PATH size mixing, attribute length validation, reserved flag bit, hold time high byte) and P3–P6 are intentionally **not** part of this PR. See `FIXPLAN.md` for the full plan.

---

Co-Authored-By: opencode <noreply@opencode.ai>


bashrusakh:
Superseded by a clean PR rebased on ruhex/main — this one pulled in unrelated fork↔upstream merge commits. Closing in favor of the new branch.

PR #5, opened 2026-06-28
fix(server): harden session lifecycle and close Cease/silent-close teardown races

bashrusakh:
## Summary

Concurrency and protocol correctness fixes for BGP session handling, plus follow-ups that close lifecycle/protocol regressions flagged in review. Six commits:

- `bd924df` — initial P1 hardening (send lock, hold timer, Cease latch, volatile state)
- `968f2e4` — split teardown reasons + close the Cease/silent-close races + atomic session removal
- `7ba07c7` — move `NotifyCeaseAsync` CAS before send; add regression test for the SilentClose-latched path
- `c58dc8b` — make session replacement atomic with `TryUpdate`; harden test reads with timeouts and IOException catch
- `5387dd3` — harden production shutdown sequencing and dispose/send teardown races
- `8b3ca2c` — keep existing Graceful Restart coverage in the branch (formatting-only test-file commit)

## Why

Correctness bugs that can cause session resets, route leaks, or protocol violations (RFC 4271 §8.1: exactly one NOTIFICATION per teardown; RFC 4724 §4: a NOTIFICATION termination bypasses Graceful Restart).

**P1 (initial commit):**
- Race on session replace — `_sessions[peerAddress] = session` clobbered the new session when the old session's finally-block ran.
- Parallel writes to `_stream` — `NetworkStream` is not thread-safe; `_sendLock` was only held in `RefreshRoutesAsync`, so keepalive / initial routes / NOTIFICATION could interleave bytes.
- Missing Hold Timer integration — the dead `BgpTimers.cs` POCO was unused; the real `HoldTimerLoopAsync` lacked the reachable `NOTIFICATION(Hold Timer Expired)` pathway.
- No NOTIFICATION on graceful termination — outer `catch (Exception)` did not emit `NOTIFICATION/Cease`.
- `_state` read without memory barriers — `IsEstablished` could observe a JIT-cached value.

**Follow-up (review-driven, commit `968f2e4`):**
- Receiving a peer NOTIFICATION returned from `ReadLoopAsync` without latching, so the `RunAsync.finally` saw `Established` + no reason and replied with a Cease — a protocol violation (RFC 4271 §6.3/§8.1: release resources/Idle, do NOT reply).
- Session replacement only fire-and-forget `NotifyCeaseAsync` — it never cancelled the old session's CTS, so the old read/keepalive loops kept running until peer close / hold timer.
- `RunSessionAsync` removal used `TryGetValue` + `TryRemove` — not atomic; a newer re-accepted session could be erased by the old finally.
- Graceful-Restart-aware shutdown skipped `NotifyCeaseAsync` but then cancelled sessions; the `RunAsync.finally` (reason still None) emitted a Cease anyway, bypassing GR (RFC 4724 §4).

**Third commit (review-driven, commit `7ba07c7`):**
- `NotifyCeaseAsync` sent Cease on the wire before CAS-latching `LocalCease`. If a concurrent `MarkSilentClose`, peer NOTIFICATION, or hold timer expiry latched a different reason between the send and the latch, the Cease already went out — violating the same RFC constraints. Fix: CAS before send, return early when reason is no longer `None`.

**Fourth commit (OCR review-driven, commit `c58dc8b`):**
- Session replacement in `AcceptLoopAsync` was not atomic: two concurrent accept threads for the same peer could both pass `TryGetValue`, both call `MarkSilentClose` on the existing session, and both execute `_sessions[peerAddress] = session` — the last writer wins, but the losing session is already started via `RunSessionAsync` and is never removed nor disposed. Fix: use `ConcurrentDictionary.TryUpdate` (atomic CAS) in a retry loop so only one replacement wins.
- `HoldTimer_Expiry` inline drain loop was missing `IOException` catch on payload reads, matching `DrainAsync` helper. Fix: added `catch (IOException)` and `catch (SocketException)` for EOF/RST robustness.
- `EstablishSessionAsync` handshake reads used `CancellationToken.None`, risking infinite hangs if the session fails to send expected messages. Fix: bounded 10-second `CancellationTokenSource`.

**Fifth commit (production hardening, commit `5387dd3`):**
- `StopAsync` now stops accepting new connections, closes the listener, and waits for the accept loop to exit before marking sessions silent-close / sending Cease. This closes the shutdown race where a late-accepted session could miss `MarkSilentClose` and later emit a protocol-incorrect Cease during GR or replacement shutdown.
- `BgpSession.Dispose()` is idempotent, and `SendMessageAsync` now treats `ObjectDisposedException` during teardown as best-effort shutdown noise instead of crashing the caller.
- Shutdown test helpers are hardened with bounded handshake reads and `SocketException`-aware draining.

## What changed

**P1 (commit `bd924df`):**
- `BGPLite.Server/BgpServer.cs` — `TryAdd` + `ReferenceEquals` guard + `NOTIFICATION/Cease` to older session on replacement conflict.
- `BGPLite.Server/BgpSession.cs` — single `SendMessageAsync` synchronized entry point; split `_advertisedPrefixesLock` from `_sendLock`; `volatile BgpFsmState _state`; `_ceaseSentOnTeardown` latch shared between `RunAsync` catch paths and the `finally` block.
- `BGPLite.Server/BgpTimers.cs` — **deleted** (dead 10-line POCO; the real `HoldTimerLoopAsync` lives next to the FSM).
- `BGPLite.Tests/BgpSessionShutdownTests.cs` — regression test `HoldTimer_Expiry_Emits_Notification_HoldTimerExpired`.

**Follow-up (commit `968f2e4`):**
- `BgpSession.cs` — replaced the bool `_ceaseSentOnTeardown` latch with a split `TeardownReason` enum (`None`/`LocalCease`/`RemoteNotification`/`HoldTimerExpired`/`SilentClose`). The `RunAsync.finally` emits a best-effort Cease ONLY on an unexpected close (reason still `None`); every other reason already produced — or deliberately suppressed — a NOTIFICATION.
  - `ReadLoopAsync` latches `RemoteNotification` on a peer NOTIFICATION before returning (no reply).
  - New `MarkSilentClose()` latches `SilentClose` and cancels the session's own CTS, so the read/keepalive loops unwind promptly.
  - All teardown-reason latch sites use `Interlocked.CompareExchange` from `None` (CAS) instead of unconditional `Exchange`, closing the race where a concurrent `MarkSilentClose` could be overwritten by a catch-block Cease (OCR review finding).
- `BgpServer.cs` —
  - Session replacement calls `MarkSilentClose()` instead of fire-and-forget `NotifyCeaseAsync` (old loops stop promptly; no Cease to the reconnecting peer).
  - `StopAsync` with GR enabled marks all sessions `SilentClose` BEFORE `_cts.Cancel()`, so the `finally` does not emit a Cease and bypass GR.
  - `RunSessionAsync` removal is now an atomic compare-and-remove via the explicit `ICollection<KeyValuePair<,>>.Remove` (removes only when key AND value match) instead of `TryGetValue`+`TryRemove`, so a newer re-accepted session is not erased.
- `BgpSessionShutdownTests.cs` — 5 new regression tests + strengthened `HoldTimer_Expiry`:
  - `Remote_Notification_Does_Not_Reply_With_Cease`
  - `MarkSilentClose_Does_Not_Send_Cease_And_Stops_Session`
  - `Replacement_MarkSilentClose_Stops_Old_Session_Promptly`
  - `Atomic_Removal_Does_Not_Erase_Replaced_Session`
  - `Unexpected_Teardown_From_Established_Emits_Exactly_One_Cease` (the `None → Cease` positive path)
  - `HoldTimer_Expiry` now asserts exactly-one-NOTIFICATION (no double Cease from the finally).

**Third commit (commit `7ba07c7`):**
- `BgpSession.cs` — `NotifyCeaseAsync` now CAS-latches `LocalCease` BEFORE `SendNotificationAsync()` and returns early when another reason was already latched. The old code sent first, then latched (post-send CAS was too late to prevent protocol-incorrect Cease in races with `MarkSilentClose`/peer NOTIFICATION/hold timer).
- `BgpSessionShutdownTests.cs` — new regression test:
  - `NotifyCeaseAsync_Does_Not_Send_When_SilentClose_Already_Latched`

**Fourth commit (commit `c58dc8b`):**
- `BgpServer.cs` — session replacement now uses `ConcurrentDictionary.TryUpdate` (atomic CAS) in a retry loop instead of `TryGetValue` + indexer set. Two concurrent accept threads for the same peer can no longer both install their session — the losing thread retries until it successfully replaces whatever is registered.
- `BgpSessionShutdownTests.cs` — `HoldTimer_Expiry` inline drain loop now catches `IOException` and `SocketException` on payload reads (matching `DrainAsync` helper). `EstablishSessionAsync` handshake reads now use a bounded 10-second `CancellationTokenSource` instead of `CancellationToken.None`.

**Fifth commit (commit `5387dd3`):**
- `BgpServer.cs` — shutdown now stops accept first, then waits for the accept loop to exit before marking sessions / cancelling session CTS. `AcceptLoopAsync` also stops on shutdown instead of spinning through errors.
- `BgpSession.cs` — `Dispose()` is idempotent and `SendMessageAsync` swallows teardown-time `ObjectDisposedException` from the send lock/stream as best-effort shutdown noise.
- `BgpSessionShutdownTests.cs` — shutdown tests now use bounded handshake reads and `SocketException`-aware draining.

**Sixth commit (commit `8b3ca2c`):**
- `GracefulRestartTests.cs` — formatting-only commit to keep the existing Graceful Restart capability/config coverage included in the PR instead of leaving it as an uncommitted local diff.

## Validation

- `dotnet build` — 0 errors, 0 warnings on changed files (1 pre-existing CS8602 in `BgpMessageTests.cs`, untouched).
- `dotnet test BGPLite.Tests` — **113/113 passed** (0 failed).
- `git diff --check` — clean.
- The entrypoint `BGPLite` project fails to build locally only because of a missing gitignored `appsettings.yml` runtime config (pre-existing, unrelated to these changes); Server and Tests build and run clean.

## Risk / regressions

- `_ceaseSentOnTeardown` (bool) replaced by `TeardownReason` (int enum) + CAS-from-None. Semantics are stricter: a concurrent silent close / peer NOTIFICATION now correctly suppresses a catch-block Cease, instead of being overwritten by it. Behavior change is limited to teardown signaling; the FSM and send path are unchanged.
- `BgpTimers.cs` deletion: confirmed no external references via grep before removal.
- Hold-timer detection granularity is `keepAliveInterval` (Math.Max(holdTime/3, 1)) so expiry is observed within `holdTime + keepalive` seconds, acceptable per RFC 4271.
- `NotifyCeaseAsync` CAS-before-send: normal path unchanged (CAS succeeds, Cease sent). Only changes behavior when another reason was already latched — then correctly returns without sending. Matches the pattern used by all other teardown paths in the same file.
- Session replacement `TryUpdate` retry loop: the loop terminates because it either wins the CAS/add or exits on shutdown/cancel; it no longer spins unbounded during teardown.
- New accept/shutdown sequencing: listener is closed and accept loop is awaited before the GR mark loop, reducing the chance of a late-accepted session missing SilentClose.
- No public API changes. No new dependencies. No behavior change outside P1 + review-fix scope.

## Out of scope

Protocol validation (path attribute order, `PrefixCodec` length guard, OPEN payload length, AS_PATH size mixing, attribute length validation, reserved flag bit, hold time high byte), routing fixes, and architectural improvements are intentionally **not** part of this PR. They will be submitted as separate, focused PRs.

---

Co-Authored-By: opencode <noreply@opencode.ai>


ruhex (approved):
Thanks for the clean rebase. I verified it against my current tree:

* builds clean;
* 113/113 tests pass;
* zero merge conflicts;
* composes correctly with the existing Graceful-Restart/Cease flow.

The new finally-Cease + MarkSilentClose implementation closes the remaining gap. I’ll run one final live accept→shutdown smoke test, but this looks good to merge.

Nice work on the TryUpdate/CAS teardown latch. Approving and merging.

ruhex:
PR #5 closes the **P1** priority (session race, send lock, hold timer, NOTIFICATION/Cease, volatile `_state`).

The remaining backlog (FIXPLAN P2–P6) is now tracked as epics with milestones:

| Epic | Milestone | Scope |
|------|-----------|-------|
| #6  | P2 — Protocol validation | attribute order, codec bounds, OPEN/AS_PATH/NEXT_HOP (RFC 4271) |
| #7  | P3 — Routing correctness | community filter, immutable Route, RouteTable races, prefix masking, LPM |
| #8  | P4 — Configuration validation | fail-loud YAML, strict keys, atomic save |
| #9  | P5 — Test coverage | parser negatives, filters, codec edges, integration |
| #10 | P6 — Architecture | FSM Active, async PeerStore, iBGP/eBGP, MaxPrefix |

Critical P2 memory-safety items (PrefixCodec length>32, OPEN length overflow, NEXT_HOP IPv6 truncation) are split into standalone `security` issues under the P2 milestone. Work proceeds P2 → P6 in that order.

Issue #12, opened 2026-06-28
[P2/security] OPEN message: optional-params/capability length silently truncated to a byte


ruhex:
## What's wrong (memory safety / malformed frame)

`BgpMessageWriter.WriteOpen` casts lengths to `byte`:
- `BGPLite.Protocol/BgpMessageWriter.cs:57` — `(byte)optParamsLen`
- `BgpMessageWriter.cs:88` — `(byte)capDataLen`
- `BgpMessageWriter.cs:93` — `(byte)cap.Data.Length`

If total optional-params / capability data exceeds **255 bytes**, the declared length is silently truncated, so the declared length no longer matches the bytes actually written → the OPEN frame is **malformed** and peers may mis-parse or crash. The optional-parameters-length field is a single byte (RFC 4271 §4.2/§6.2).

## Fix
Validate `<= 255` and reject (or split) an OPEN whose capability data overflows the single-byte length field, instead of truncating.

## Acceptance
- An OPEN whose total optional-params/capability data > 255 bytes is rejected, not silently length-truncated.
- Unit test covers the overflow case.

Part of epic #6.

bashrusakh:
Closed by https://github.com/ruhex/BGPLite/pull/38

Fix landed in `fix/p2-overflow-guards`:

- `BgpMessageWriter.WriteOpen` and `WriteCapabilities` no longer silently cast `(byte)` on the single-byte length fields. A new `RequireFitsByte` helper validates the values up front and throws `ArgumentOutOfRangeException` on overflow.
- Three length fields are now guarded: total `optParamsLen`, total `capDataLen`, and per-capability `Data.Length`.
- Tests: `Open_SingleCapabilityExceedingByte_Throws`, `Open_TotalCapabilityDataExceedingByte_Throws`, and `Open_AtByteBoundary_Succeeds` covering the 255 boundary.

One note for review: any single capability with `Data.Length > 255` also pushes the surrounding `optParamsLen` past 255, so the `WriteOpen` guard always fires first. The per-capability check in `WriteCapabilities` is still useful as defense-in-depth and as a clearer error message in the (impossible-to-reach-via-Encode) direct case, but if you'd prefer to drop it to keep the code minimal, that's fine too.

Four rounds of OpenCodeReview were run; every actionable comment was applied. `dotnet test`: 121/121 passed.

ruhex:
[...]

PR #38, opened 2026-07-02
protocol: harden PrefixCodec and OPEN overflow guards (closes #11, #12)

### LINKED ISSUE #11: [P2/security] PrefixCodec: out-of-bounds read/write for prefix length > 32
## What's wrong (memory safety)

`PrefixCodec` has no guard against a prefix `length > 32`. For such a length the encode loop runs `i >= 4` iterations:

- Encode: `BGPLite.Protocol/PrefixCodec.cs:20-21` — `buffer[1+i] = (byte)(addr >> (24 - i*8))`. `24 - i*8` becomes a **negative shift** and `buffer[1+i]` writes **past the 4-byte IPv4 address** into the caller's span.
- Decode mirrors it: `PrefixCodec.cs:35` (`addr |= ... << (24 - i*8)`) and `:37` (`addr &= 0xFFFFFFFF << (32 - length)` — negative shift for `length > 32`).

The `IpPrefix.Address` field is `uint` / IPv4-only, so `length` must be in `0..32` (RFC 4271 §4.3).

## Fix
Reject `length > 32` with `ArgumentOutOfRangeException` (or clamp) in both `Encode` and `Decode`.

## Acceptance
- `PrefixCodec.Encode`/`Decode` reject `length == 33` instead of writing OOB.
- Unit test covers the `length > 32` case.

Part of epic #6.

bashrusakh:
## Summary

Closes #11 and #12.

- **#11** `PrefixCodec` OOB read/write for IPv4 prefix length > 32
- **#12** `BgpMessageWriter.WriteOpen` silently truncates single-byte length fields when capability data exceeds 255 bytes

Both are memory-safety / malformed-frame bugs in the protocol layer. This PR lands the smallest correct fix at the protocol boundary (single point of truth for encoding) and hardens the buffer-size checks around the same area.

## Why

RFC 4271 §4.2 declares the optional-parameters length as a single byte, and §6.2 does the same for capability data lengths. The previous code cast these to `byte` after the length was computed, which silently truncated any value > 255 and produced malformed frames. Similarly, `PrefixCodec` performed negative bit shifts and out-of-bounds writes for `length > 32`, which is undefined for an IPv4 address. `WriteHeader` cast `int totalLength` to `(ushort)` with no bounds check, silently truncating any value > 65535. The OPEN length guards fired AFTER the caller's buffer was already partially mutated.

## Changes

- **`PrefixCodec.Encode` / `Decode`**: reject `length > 32` with `ArgumentOutOfRangeException`.
- **`PrefixCodec.Encode` / `Decode`**: reject empty buffer and reject buffer too small for the declared prefix length.
- **`BgpMessageWriter.WriteOpen` / `WriteCapabilities`**: validate `optParamsLen`, `capDataLen`, and per-capability `Data.Length` against `byte.MaxValue` before casting, via a new `RequireFitsByte` helper. Throws `ArgumentOutOfRangeException` instead of silently truncating.
- **`BgpMessageWriter.WriteOpen`**: all length checks now run BEFORE `WriteHeader` and any buffer writes, so a rejected write leaves the caller's `Span<byte>` untouched. `WriteCapabilities` keeps its checks as defense-in-depth for the direct-call path.
- **`BgpMessageWriter.WriteHeader`**: reject `totalLength` outside `19..ushort.MaxValue` before the `(ushort)` cast. The 19 floor matches the BGP header size; the ceiling is what fits in the 16-bit length field per RFC 4271 §4.1. (We don't enforce a 4096 hard cap here — legitimate UPDATEs can exceed it.)
- **`GetCapabilitiesDataLength`** (new helper): single source of truth for the `capDataLen` computation, replacing a duplicated loop in `GetOptParamsLength`.

## Commits

```
4b364a2 protocol: validate OPEN lengths before mutating buffer; cap WriteHeader
09deaa3 protocol+tests: add symmetric buffer-size guard to PrefixCodec.Decode
b2eb840 protocol+tests: harden PrefixCodec buffer guards (OCR round 2)
5746269 tests: clarify which OPEN overflow path each test exercises
13c9687 protocol: reject OPEN with optional-params or capability data > 255 bytes
f66d8d7 protocol: reject IPv4 prefix length > 32 in PrefixCodec
```

Five rounds of OpenCodeReview were run against the branch; every actionable comment was applied.

## Validation

- `dotnet test --nologo`: **139/139 passed** on `ruhex/main` base (+17 new tests vs base).
- New tests:
  - `PrefixCodecTests`: `Encode_LengthAbove32_Throws` (×4), `Decode_LengthAbove32_Throws` (×3), `Encode_DoesNotWriteBeyondBuffer_ForValidPrefix`, `Encode_EmptyBuffer_Throws`, `Encode_BufferTooSmallForPrefix_Throws`, `Decode_EmptyBuffer_Throws`, `Decode_BufferTooSmallForPrefix_Throws`.
  - `BgpMessageTests`: `Open_SingleCapabilityExceedingByte_Throws`, `Open_TotalCapabilityDataExceedingByte_Throws`, `Open_AtByteBoundary_Succeeds` (255 boundary), `Open_Overflow_DoesNotMutateBuffer` (sentinel pattern), `Open_TotalOverflow_DoesNotMutateBuffer` (sentinel pattern).
- Diff scope: **4 files, +259/-7**, all in `BGPLite.Protocol/` and `BGPLite.Tests/`. Branch is rebased on `ruhex/main` (`989c116`) — no drift, no scope creep.

## Risk

Low. Changes are fail-loud in the protocol boundary on previously-unreachable or out-of-band inputs:
- Callers that only pass valid IPv4 prefixes (length 0..32) and well-formed capabilities are unaffected.
- Callers that fed `length > 32`, oversized capability data, or `totalLength > 65535` were already producing undefined/malformed output; they now get a clear exception at the point of misuse.
- The new atomic-write guarantee means a failed validation no longer leaves the caller's `Span<byte>` partially mutated; the test suite asserts this with a sentinel-pattern.

No public API changes, no DB schema changes, no runtime config changes, no new dependencies.

## Checklist

- [x] Diff vs `ruhex/main` matches scope of #11 and #12
- [x] Focused tests added (encode, decode, boundary, overflow, atomic-write)
- [x] Full `dotnet test` green on rebased branch
- [x] No new dependencies
- [x] No public API / schema / config changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stronger validation for BGP message sizes and optional parameter lengths, with clear errors when values exceed allowed limits.
  * Improved prefix encoding/decoding safety by rejecting invalid lengths and preventing reads or writes past buffer bounds.

* **Tests**
  * Added regression coverage for boundary cases, overflow handling, and buffer safety.
  * Verified valid edge cases still succeed, including exact byte-limit scenarios.

### DIFF
--- a/BGPLite.Protocol/BgpMessageWriter.cs
+++ b/BGPLite.Protocol/BgpMessageWriter.cs
@@ -54,6 +54,9 @@ public static class BgpMessageWriter
         p += 4;
 
         var optParamsLen = GetOptParamsLength(msg.Capabilities);
+        // RFC 4271 §4.2: optional-parameters length is a single byte.
+        // Silently truncating > 255 yields a malformed frame; reject up front.
+        RequireFitsByte(optParamsLen, nameof(BgpOpenMessage.Capabilities), "optional-parameters");
         buffer[p++] = (byte)optParamsLen;
 
         WriteCapabilities(msg.Capabilities, buffer[p..]);
@@ -85,15 +88,25 @@ public static class BgpMessageWriter
 
         var p = 0;
         buffer[p++] = 2; // Type: Capabilities
+        // RFC 4271 §6.2: capability data length is a single byte.
+        RequireFitsByte(capDataLen, nameof(BgpCapabilityInfo.Data), "capability data");
         buffer[p++] = (byte)capDataLen;
 
         foreach (var cap in capabilities)
         {
             buffer[p++] = cap.Code;
+            // RFC 4271 §6.2: per-capability length is a single byte.
+            RequireFitsByte(cap.Data.Length, nameof(BgpCapabilityInfo.Data), "capability length");
             buffer[p++] = (byte)cap.Data.Length;
             cap.Data.AsSpan().CopyTo(buffer[p..]);
             p += cap.Data.Length;
         }
+    }
+
+    private static void RequireFitsByte(int value, string paramName, string field)
+    {
+        if (value < 0 || value > byte.MaxValue)
+            throw new ArgumentOutOfRangeException(paramName, value, $"{field} length must fit in a single byte (0..{byte.MaxValue}).");
     }
 
     #endregion
--- a/BGPLite.Protocol/BgpMessageWriter.cs
+++ b/BGPLite.Protocol/BgpMessageWriter.cs
@@ -30,6 +30,12 @@ public static class BgpMessageWriter
 
     private static void WriteHeader(BgpMessageType type, int totalLength, Span<byte> buffer)
     {
+        // RFC 4271 §4.1: BGP message length is a 16-bit field, min 19 (header only).
+        // Reject any length outside the representable range up front to avoid a
+        // silent (ushort) truncation that would emit a malformed frame.
+        if (totalLength < 19 || totalLength > ushort.MaxValue)
+            throw new ArgumentOutOfRangeException(nameof(totalLength), totalLength, $"BGP message length must be in 19..{ushort.MaxValue}.");
+
         BgpConstants.Marker.CopyTo(buffer);
         BinaryPrimitives.WriteUInt16BigEndian(buffer[16..], (ushort)totalLength);
         buffer[18] = (byte)type;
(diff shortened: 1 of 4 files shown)


bashrusakh:
@coderabbit review

Issue #23, opened 2026-07-01
[P3/api] /api/me resolves the peer by client IP — ambiguous when several peers share a source IP


ruhex:
## Summary

`/api/me` (`HandleGetMe`) resolves the caller's peer record by the HTTP client IP via `GetPeerByIp`. When several peers sit behind one NAT/VPN and share the API's client IP, the endpoint returns the **first** match and cannot tell the peers apart.

This is the API-layer follow-up of the same root cause as #18 / #19 (IP-only peer identity).

## Root cause

- `BGPLite.Api/ManagementApi.HandleGetMe`: `var peerInfo = _store.GetPeerByIp(clientIp);`
- `PeerStore.GetPeerByIp` is Ip-only by design (kept in #19 for backward compatibility).
- The HTTP request carries no AS, so the endpoint has nothing to disambiguate on.

## Expected behaviour

A peer querying `/api/me` from behind a shared source IP gets **its own** record (the one matching its AS), not another peer's.

## Proposed fix

Let the caller disambiguate. Options:

- Accept the peer's AS on the call (e.g. `?asn=` query param or a header) and resolve via `GetPeer(ip, asn)` (added in #19); **or**
- when more than one peer matches the IP, return a list so the caller can pick; keep the single-object response when exactly one matches (backward compatible).

Low risk; touches only `ManagementApi.HandleGetMe` (+ request model if a param is added). Add a test covering two peers at one IP.

Refs #18, #19.

Issue #22, opened 2026-07-01
[P2/routing] Community route filter is Ip-scoped — peers behind one source IP share one allow-list


ruhex:
## Summary

`PeerCommunityFilter` resolves a peer's community allow-list with `GetCommunitiesByIp(ip)`, which **merges** the communities of every peer sharing that source IP. After #19 each such peer is its own `(Ip, Asn)` row, so the merged allow-list is wrong: a route tagged with one peer's community can be sent to the other peers at the same IP, and vice-versa.

This is the routing-layer follow-up of the same root cause as #18 / #19 (IP-only peer identity).

## Root cause

- `BGPLite/Program.cs`: `new PeerCommunityFilter(ip => store.GetCommunitiesByIp(ip))`.
- `BGPLite.Routing/PeerCommunityFilter.AcceptOutgoing(Route, PeerConfig)` only has `peer.Address` (the IP).
- `PeerConfig.RemoteAsn` is **not populated** for dynamically accepted peers — the AS is known only after OPEN (`BgpSession._remoteAsn`), while `PeerConfig` is built at accept time and its setters are `init`.

So the filter cannot disambiguate peers behind one source IP.

## Expected behaviour

Community filtering is per peer: a route is sent to peer `(Ip, Asn)` only if it matches **that** peer's community set (stored per `PeerCommunity.PeerId`, i.e. per `(Ip, Asn)` row after #19).

## Proposed fix

Thread the session's effective remote AS into the filter:

- Populate the session's AS so the filter can see it — e.g. set `PeerConfig.RemoteAsn` (make it settable / add a runtime field) in `BgpSession` after OPEN, or change `IRouteFilter.AcceptOutgoing` to take the ASN.
- Add `PeerStore.GetCommunities(string ip, uint asn)` (per `(Ip, Asn)`) and have `PeerCommunityFilter` use it.
- Keep `GetCommunitiesByIp` for backward compatibility if needed elsewhere, or remove it once all callers move to `(Ip, Asn)`.

## Risk / scope

Touches the `BGPLite.Routing` boundary and the `IRouteFilter` contract — slightly wider than #18/#19. No protocol impact; affects only the (optional) community-tagged outgoing route selection. Add tests covering two peers from one IP with disjoint community sets.

Refs #18, #19.

ruhex:
[...]

PR #77, opened 2026-07-03
fix(api): key peer identity by (Ip, Asn)

bashrusakh:
## Summary
- Make shared-IP peers resolve by (Ip, Asn) instead of IP-only.
- Disambiguate `/api/me` with `?asn=` and return 409 when the caller is ambiguous.
- Add regression coverage for shared-IP peer storage and API behavior.

## Why
- NAT/shared-IP neighbors were leaking into the wrong peer record / community view.
- The API needed an explicit way to identify the correct peer without exposing another peer's state.

## Validation
- `dotnet format BGPLite.sln --verify-no-changes`
- `dotnet test BGPLite.Tests/BGPLite.Tests.csproj --no-restore`

## Risk
- Backports the peer identity model to composite (Ip, Asn) and updates the API contract for ambiguous `/api/me` responses.

## Summary by CodeRabbit

* **New Features**
  * The management API “me” endpoint now supports peers that share the same IP address and can return the correct peer when paired with `?asn=`.
  * Added support for listing all peers associated with a given source IP.

* **Bug Fixes**
  * Improved peer selection for shared-IP scenarios, including clearer `400` (invalid ASN) and `409` (ambiguous) responses.
  * Communities are now returned based on the selected peer record (not aggregated by IP alone).

* **Documentation**
  * Updated the README for the “me” endpoint and the new `?asn` behavior.

### DIFF
--- a/BGPLite.Api/BgpDbContext.cs
+++ b/BGPLite.Api/BgpDbContext.cs
@@ -26,7 +26,6 @@ public class BgpDbContext : DbContext
         db.Database.ExecuteSqlRaw("DROP INDEX IF EXISTS IX_Peers_Ip;");
         db.Database.ExecuteSqlRaw(
             "CREATE UNIQUE INDEX IF NOT EXISTS UX_Peers_Ip_Asn ON Peers (Ip, Asn);");
-
         db.Peers.Where(p => p.Status == "active").ExecuteUpdate(
             s => s.SetProperty(p => p.Status, "inactive"));
     }
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -359,7 +359,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         var subscriptions = _store.GetSubscriptions(peer.Id);
         var customPrefixes = _store.GetCustomPrefixes(peer.Id);
         var customAsns = _store.GetCustomAsns(peer.Id);
-        var communities = _store.GetCommunitiesByIp(peer.Ip);
+        var communities = _store.GetCommunities(peer.Id);
 
         return ApiResponse.Ok(new
         {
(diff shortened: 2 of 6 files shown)


bashrusakh:
Addressed in the latest push (dc9dd90):

- `HandleGetPeer` now uses `peer.Id` for communities, so shared-IP peers no longer leak filters across identities.
- `GetClientIp()` now uses the request `RemoteEndPoint` only; it no longer trusts `X-Real-IP` / `X-Forwarded-For`.
- `GetPeersByIp()` now reuses `MapToInfo()`.
- Removed the duplicate `BGPLite.Api` project reference from `BGPLite.Tests.csproj`.

Verified with:
- `dotnet format BGPLite.sln --verify-no-changes`
- `dotnet test BGPLite.Tests/BGPLite.Tests.csproj --no-restore` (207/207)

If the bot comment is still visible, this review is already fixed in the latest push.

ruhex:
Closing — this regresses `/api/me` for reverse-proxied deployments.

The removal of `X-Real-IP` / `X-Forwarded-For` handling in `GetClientIp` means that when the management API sits behind a reverse proxy (our production setup — nginx), `/api/me` sees the **proxy's** IP via `RemoteEndPoint.Address` instead of the real client. It then resolves no peer → 404, breaking the endpoint's core contract. Honoring `X-Forwarded-For` / `X-Real-IP` behind an L7 proxy is standard practice (nginx/Cloudflare/etc.), so this isn't optional for us.

The (Ip, Asn) peer-identity model and the `?asn=` disambiguation + 409 are good — please re-submit **keeping the `X-Forwarded-For` / `X-Real-IP` resolution** (as the inner step before `GetPeersByIp`, so the proxy IP is unwrapped first and *then* disambiguated by ASN). Happy to re-review that version.

Refs #22, #23.

Issue #19, opened 2026-07-01
[P0/api] PeerStore keyed by remote IP only — peers behind one source IP share one record (composite Ip+Asn key)


ruhex:
## Summary

The peer table has a **UNIQUE index on `Ip` alone**, and every PeerStore lookup (`CreatePeer`, `UpsertPeer`, `UpdateSessionStatus`, `GetPeerByIp`) matches by IP only. Distinct peers that connect from the same source IP (e.g. several routers behind one NAT/VPN, different AS) therefore collapse into a single peer record: the second connection overwrites the first's AS, and both share one set of subscriptions / custom prefixes / custom ASNs / status.

This is the **persistence-layer half** of the same root cause as #18 (which fixed the in-memory session map). With #18 the sessions no longer drop, but each session still resolves to the wrong peer record unless this is fixed too.

## Impact

Correctness. Two peers behind one source IP are served the other's configured prefixes/subscriptions; their status flips between them; and the unique `Ip` index makes it impossible to represent them as separate peers at all.

## Root cause

- `BGPLite.Api/BgpDbContext.cs` — `e.HasIndex(p => p.Ip).IsUnique();` enforces one row per IP.
- `BGPLite.Api/PeerStore.cs` — `CreatePeer` / `UpsertPeer` / `UpdateSessionStatus` all do `FirstOrDefault(p => p.Ip == ip)`.
- `BgpSession.SendAllRoutesAsync` resolves the peer with `GetPeerByIp(_peerConfig.Address)` (IP only), even though the session already knows `_remoteAsn`.

## What the RFC says

The AS is the natural per-peer identity alongside the address: per RFC 4271 §4.2 the OPEN *"My Autonomous System"* identifies the sender, and §6.2 leaves acceptable AS numbers to local policy. Real BGP daemons (FRR / BIRD / OpenBGPD) model a configured neighbor as `(address, ASN)`. (The BGP Identifier, RFC 4271 §4.2, is an even more stable per-device discriminator and is intentionally left as a future refinement.)

## Expected behaviour

Two peers from one source IP with different AS are separate peer rows with independent subscriptions / prefixes / status, and each BGP session resolves to its own `(Ip, Asn)` record.

## Proposed fix

Make the durable peer identity `(Ip, Asn)`:

[...]

PR #78, opened 2026-07-03
fix(api): restore reverse-proxy IP resolution for /api/me

### LINKED ISSUE #23: [P3/api] /api/me resolves the peer by client IP — ambiguous when several peers share a source IP

bashrusakh:
## Summary
- Restore `X-Real-IP` / `X-Forwarded-For` resolution in `/api/me` before peer lookup.
- Keep `?asn=` disambiguation for shared-IP peers.
- Add regression tests for client-IP resolution.

## Why
- Reverse-proxied deployments (nginx) were seeing the proxy IP instead of the real client.

## Validation
- `dotnet test BGPLite.Tests/BGPLite.Tests.csproj --no-restore`
- `dotnet format BGPLite.sln --verify-no-changes`

## Risk
- `/api/me` again trusts standard reverse-proxy headers when present, matching the prior behavior.

Closes #23

## Summary by CodeRabbit

* **New Features**
  * Improved peer lookup for the “me” endpoint, including support for identifying the correct peer when multiple peers share an IP by using an ASN filter.
  * Client IP detection is now more reliable behind proxies, using forwarded headers before falling back to the remote address.

* **Bug Fixes**
  * Communities are now returned from the correct peer record, ensuring peer-specific data is shown accurately.
  * Peer records are now treated as unique per IP and ASN combination, reducing ambiguity and duplicate matches.

### DIFF
--- a/BGPLite.Api/BGPLite.Api.csproj
+++ b/BGPLite.Api/BGPLite.Api.csproj
@@ -17,4 +17,8 @@
     <ProjectReference Include="..\BGPLite.Server\BGPLite.Server.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="BGPLite.Tests" />
+  </ItemGroup>
+
 </Project>
--- a/BGPLite.Api/BgpDbContext.cs
+++ b/BGPLite.Api/BgpDbContext.cs
@@ -26,7 +26,6 @@ public class BgpDbContext : DbContext
         db.Database.ExecuteSqlRaw("DROP INDEX IF EXISTS IX_Peers_Ip;");
         db.Database.ExecuteSqlRaw(
             "CREATE UNIQUE INDEX IF NOT EXISTS UX_Peers_Ip_Asn ON Peers (Ip, Asn);");
-
         db.Peers.Where(p => p.Status == "active").ExecuteUpdate(
             s => s.SetProperty(p => p.Status, "inactive"));
     }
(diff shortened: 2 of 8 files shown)


ruhex:
Closing — superseded by this session's work. Most of PR #78 already landed through merged PRs:

- **Composite (Ip, Asn) key** — merged in #19 (schema) + #148 (PeerStore CRUD)
- **Trusted-proxy client-IP resolution** — merged in #91 (ResolveClientIp + XFF walk + TrustedProxies gate)
- **GetPeer(ip, asn)** — exists in PeerStore since #19

The only remaining piece (the `?asn=` disambiguation query param on /api/me) will ship in a fresh focused PR that closes #23.

Issue #151, opened 2026-07-05
[bug/providers] PrefixSourceUrlValidator ConnectCallback is IPv4-only — IPv6-first hosts throw SocketException


ruhex:
## Bug

`PrefixSourceUrlValidator.CreateValidatedConnectionAsync` (the `SocketsHttpHandler.ConnectCallback` that backs the SSRF defense from #144) is **IPv4-only** and throws on IPv6-first / IPv6-only hosts.

## Root cause

`PrefixSourceUrlValidator.cs:71`:
```csharp
var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp) { NoDelay = true };
await socket.ConnectAsync(addresses[0], port, ct);   // :74 — addresses[0] may be IPv6
```

`Dns.GetHostAddressesAsync` returns addresses in OS order; for many hosts `addresses[0]` is IPv6. The validator loop (`:64-69`) correctly validates **all** resolved IPs against the blocklist (which includes IPv6 ranges at `:28-31`), but the connect hardcodes an `InterNetwork` (IPv4) socket — handing it an IPv6 `addresses[0]` raises `SocketException` (`AddressFamilyNotSupported`). There is no fallback to other resolved addresses.

## Impact

- Pre-existing, surfaced wider by #147: peers' user-supplied URLs hit arbitrary public hosts (IPv6-only CDNs are common), so user-source fetches will silently fail (caught per-source, logged, source skipped) for IPv6-first targets.
- Operator `PrefixSources` are less affected (operator controls the URLs), but the same bug applies.

## Fix

Match the socket `AddressFamily` to the chosen address, and prefer connecting to the first reachable address (try addresses in order, or pick the family the OS route table would). Keep the SSRF validation of **all** addresses before connecting to any.

## Scope

S–M. Pure provider-layer fix; covered by extending `PrefixSourceUrlValidatorTests`.

Refs: #144, #147.

Issue #147, opened 2026-07-05
[143-2] SendAllRoutesAsync: fetch per-peer user-sources + community stamp


ruhex:
Part of #143 (epic). Second sub-task: integration into the send path.

Depends on: #143-1 (entity + store).

## What
For each peer in SendAllRoutesAsync, after CustomPrefixes/CustomAsns, fetch the peer's custom-sources (URLs) via HttpPrefixProvider and add to the route list.

### SendAllRoutesAsync
```
// After custom prefixes/custom ASNs:
var customSources = _peerStore.GetCustomSources(peer.Id);
foreach (var source in customSources)
{
    var sourcePrefixes = await _httpPrefixProvider.LoadAsync(new PrefixSourceConfig {
        Kind = "http", Name = source.Name, Url = source.Url,
        Community = source.Community, Timeout = 30
    }, _cts.Token);
    var sourceComms = source.Community is not null
        ? ParseCommunity(source.Community)
        : _communityResolver.Resolve(new CommunitySource(CommunitySourceKind.UserSource, source.Name));
    foreach (var (prefix, length) in sourcePrefixes)
        routes.Add(MakeRoute(prefix, length, nextHop, null, sourceComms));
}
```

### Community
- If source.Community is specified → parse "ASN:VALUE" → stamp.
- If not → auto from 5XX range (hash of name or reserved kind in ConfigCommunityResolver).
- New `CommunitySourceKind.UserSource` in the enum.

### Per-peer scoping
- Custom-source prefixes are added to THIS peer's route list only (same as CustomPrefixes).
- NOT in the shared RouteTable. Never cross-contaminate between peers.
- PeerCommunityFilter applies on the resolved allow-set as usual.

### Tests
- SendAllRoutesAsync: mock HttpPrefixProvider returns prefixes → routes added with correct community.
- Community stamping: specified community vs auto-generated.

Issue #144, opened 2026-07-05
[security] SSRF defense: validate fetch URLs + cap response size (#143 prerequisite)


ruhex:
Hard prerequisite for #143 (user-supplied prefix-list URLs).

## Design: ConnectCallback (NOT pre-resolve)

SSRF validation at the **socket level** via `SocketsHttpHandler.ConnectCallback` — DNS resolution + IP validation happen ONCE, on the IP that is actually connected to. No TOCTOU race.

### Defenses
1. **SSRF**: `ConnectCallback` resolves DNS, checks each IP against BlockedRanges (private RFC 1918, loopback, link-local incl. cloud metadata, CGNAT, benchmarking, multicast, reserved, IPv6 ULA/loopback/link-local), connects only if ALL resolved IPs are public.
2. **Redirect bypass**: `SocketsHttpHandler` does NOT follow redirects by default (unlike `HttpClientHandler`). No 302-to-internal-IP bypass.
3. **OOM**: HttpPrefixProvider streams response with 10MB cap (ResponseHeadersRead + manual size check).

### What was removed (from initial approach)
- `dnsResolver` parameter on HttpPrefixProvider (was a testability workaround for pre-resolve TOCTOU design).
- Pre-resolve `ValidateUrlAsync` call in HttpPrefixProvider.LoadAsync (moved to handler-level ConnectCallback).
- `ValidateUrlAsync` + tests retained for future API-level URL validation (when user submits URL via API).

## Tests
- `IsBlockedAddress`: 13+ blocked IP ranges + public IPs pass.
- `ValidateUrlAsync`: mocked-DNS integration tests (for future API use).
- HttpPrefixProvider tests: no DNS dependency (StubHandler bypasses ConnectCallback).

Issue #143, opened 2026-07-05
[epic] User-supplied prefix-list URLs — peers bring their own CIDR sources


ruhex:
## Idea
Let peers (via the management API) add their OWN URL-based prefix-list sources (CIDR-per-line files) alongside the existing operator-configured PrefixSources and per-peer CustomPrefixes (CIDR strings). Turns BGPLite from a one-way feed into a collaboration platform where peers bring their own lists.

**Existing features stay as-is:** operator PrefixSources (YAML), per-peer CustomPrefixes (CIDR strings), per-peer Subscriptions (ASN-list names). This is a NEW, additional source kind per peer.

## Architecture: per-peer scoping (NOT shared RouteTable)

User-supplied URL sources are handled **exactly like CustomPrefixes / CustomAsns** — fetched per-peer in `SendAllRoutesAsync`, added to THAT peer's route list, advertised ONLY to that peer. They do NOT go into the shared `RouteTable`.

```
SendAllRoutesAsync (for Peer A):
  1. Shared RouteTable (Cloudflare, Google, ...)              ← shared, all peers
  2. CustomPrefixes (CIDR strings, Peer A)                    ← ONLY A
  3. CustomAsns (ASN lists, Peer A)                           ← ONLY A
  4. UserSources (URL CIDR lists, Peer A)  ← NEW              ← ONLY A
  → advertise TO Peer A ONLY
```

This means: Peer A's user-source prefixes NEVER reach Peer B, even if both use the same community value. Isolation is guaranteed by the per-peer send path, not by community uniqueness.

### Community design

Community is a **wire tag for the receiving peer's filtering**, NOT an access-control mechanism. Multiple peers can use the same community value (e.g. `65444:501`) for different URL sources — their prefixes never cross-contaminate because they're in per-peer route lists.

When adding a source via API, the user can optionally specify a community:
```json
POST /api/peers/{id}/sources
{
  "name": "my-blocklist",
  "url": "https://example.com/blocklist.txt",
  "community": "65444:501"    // optional; if omitted → auto from 5XX range
}
```

- If specified → stamped on every prefix from that source (like `PrefixSourceConfig.Community`).
- If omitted → auto-generated from a reserved range (e.g. `65444:500 + hash(name) % 100`).
- No uniqueness enforcement across peers — duplicates are safe (per-peer scoping guarantees isolation).
- `ExactUnionPrefixAggregator` groups by community set within each peer's route list → correct aggregation.

## What to add

[...]

PR #152, opened 2026-07-05
perf(providers): URL-keyed TTL cache for per-peer user-source fetches (#150)

### LINKED ISSUE #150: [perf/providers] Cache per-peer user-source URL fetches (TTL, URL-keyed)
## Context

Follow-up to #147 (epic #143). Per-peer user-supplied URL sources are now fetched and advertised, but **every active source URL is re-fetched on every `SendAllRoutesAsync` refresh, for every peer** — there is no dedup or TTL cache on the per-peer path.

## Why

The existing `PrefixSourceService` cache (`PrefixSourceService.cs`) is keyed by **operator `PrefixSources` name** (YAML config) — it only covers sources in `AppConfig.PrefixSources`. Per-peer `PeerCustomSource` rows live in SQLite and are not in that collection, so `HttpPrefixProvider.LoadAsync` is hit cold on every send. With N peers × M active sources × refresh interval, that is N×M HTTP fetches per refresh.

`HttpPrefixProvider` itself is stateless by design (safe as a singleton) — the cache belongs one layer up.

## Proposal

A URL-keyed in-process cache (with TTL + negative-TTL, mirroring `PrefixSourceService`'s shape) wrapping the per-peer user-source fetch — either:
- inside `PrefixService.GetUserSourcePrefixesAsync` (key = url), or
- a small dedicated cache collaborator injected alongside `HttpPrefixProvider`.

Keying by URL (not by source name) dedupes across peers that happen to subscribe to the same list. Stale-on-failure serving (like `PrefixSourceService`) keeps a peer's routes stable through transient fetch errors.

## Scope

M. No behavior change for operator `PrefixSources` (already cached). Respects the #114 cancellation rule and the #144 SSRF defense (cache sits above the provider, transparent to it).

Refs: #147, #143.

ruhex:
## What

Closes #150 — follow-up to #147 (epic #143). Adds a URL-keyed in-memory TTL cache for per-peer user-supplied prefix-list sources, so they're no longer cold-fetched on every `SendAllRoutesAsync` refresh for every peer.

## Why

After #147, each active user source was re-fetched on every refresh, per peer. The existing `PrefixSourceService` cache couldn't help — it's keyed by **operator** `PrefixSources.Name` and peer URLs aren't in `AppConfig.PrefixSources`. With N peers × M active sources × refresh interval, that was N×M cold HTTP fetches per refresh. `HttpPrefixProvider` is stateless by design, so the cache belongs one layer up.

## How

New `UserSourceCache` (`internal`, `BGPLite.Providers`), mirroring `PrefixSourceService`'s proven shape:
- **URL-keyed** `ConcurrentDictionary` + per-key `SemaphoreSlim` → no thundering herd on a cold/expired key.
- **Positive TTL 1h, negative TTL 30s** (injectable for tests).
- **Stale-on-failure**: a transient fetch error serves the last good copy, so a peer's routes stay stable; a cold failure (no cached copy) records a short-lived negative entry so repeated calls don't hammer the upstream.
- **OCE propagates** (#114) and is never recorded as negative.
- **Logs by source `Name`** (the safe identifier), **never the URL** — peer URLs may carry query-string tokens (#149 redaction preserved).

`PrefixService.GetUserSourcePrefixesAsync` now goes through the cache. SSRF defense (#144) is inherited — the cache sits above `HttpPrefixProvider`, transparent to it. No `BgpSession` change (the seam is unchanged).

### Why URL-keyed (shared across peers)

Keying by URL — not by `(peer, source)` or source name — means peers that subscribe to the **same list share one fetch**. This is the common case (e.g. several peers on the same popular blocklist) and it's the main memory/traffic win.

### Paused / disabled sources (explicit)

The cache knows nothing about a source's `Active` flag — by design. The pause lifecycle is handled **above** the cache:
- `LoadPeerRoutingView` applies `Where(c => c.Active)` (`PeerStore.cs`) before the send path reaches the cache, so a **paused source contributes no prefixes regardless of cache state** — it's "filtered in advance", not evicted.
- Because the cache is shared (URL-keyed), pausing or deleting a source in one peer does **not** evict the entry — another peer may still need the same URL. Orphaned entries (no active subscriber left) simply expire via TTL.

So: disable = "not considered during prefix formation" (immediate, next refresh); the shared entry stays for any other subscriber and ages out otherwise. No cross-layer invalidation coupling needed.

## Verification

- `dotnet build` — 0 errors.
- `dotnet test` — **407/407** (7 new in `UserSourceCacheTests`: same-URL dedup, distinct-URL separation, stale-on-failure, cold-failure negative-cache, OCE propagation + not-negative-cached, positive/negative TTL expiry).
- `dotnet format` — clean.
- Operator `PrefixSources` path unchanged (still uses `PrefixSourceService`'s name-keyed cache).

## Follow-ups

- **#151** (IPv4-only `Socket` in `PrefixSourceUrlValidator`) remains open — separate concern.

Refs: #150, #147, #143.

## Summary by CodeRabbit

* **New Features**
  * Improved per-URL caching and de-duplication for user-provided prefix-list sources, reducing repeated remote fetches.
  * Added controlled retry behavior with TTLs to keep prefix data fresh without unnecessary reloads.
* **Bug Fixes**
  * If an update fails, the last known good prefix list is served instead of breaking responses.
  * Cancellations now propagate correctly and do not poison the cache; failures are briefly negative-cached.
* **Tests**
  * Expanded test coverage for TTL, concurrency, failure, cancellation, and per-URL isolation.

### DIFF
--- a/BGPLite.Providers/PrefixService.cs
+++ b/BGPLite.Providers/PrefixService.cs
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using BGPLite.Configuration;
+using Microsoft.Extensions.Logging;
 
 namespace BGPLite.Providers;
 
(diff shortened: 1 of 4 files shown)


ruhex:
## CodeRabbit review — addressed

CodeQL/`analyze`: **clean** (0 open alerts). CodeRabbit: 1 nitpick (test coverage), no actionable.

**Fixed (451f295):**
- 📐 *Nitpick — concurrent thundering-herd test* (`UserSourceCacheTests`): the sequential dedup test only exercised cache reuse, not the per-URL `SemaphoreSlim` gate. Added `Concurrent_Same_Url_Fetched_Once_Gate_Serializes` — 16 concurrent racers with a blocking fetcher, asserts exactly one fetch (the actual gate behavior). Verified non-flaky (3 consecutive runs).

Build/test/format re-verified green (408/408).

PR #153, opened 2026-07-05
fix(providers): connect SSRF-validated hosts IPv4-first, fall through on failure (#151)

### LINKED ISSUE #151: [bug/providers] PrefixSourceUrlValidator ConnectCallback is IPv4-only — IPv6-first hosts throw SocketException

ruhex:
## What

Closes #151 — fixes the SSRF `ConnectCallback` so it no longer throws on hosts whose first DNS record is IPv6, and actually connects on IPv4-only service hosts.

## Bug

`PrefixSourceUrlValidator.CreateValidatedConnectionAsync` (#144) hardcoded an IPv4 socket but connected to `addresses[0]`:
```csharp
var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp) { NoDelay = true };
await socket.ConnectAsync(addresses[0], port, ct);   // addresses[0] may be IPv6
```
DNS often returns IPv6 first on dual-stack hosts → `SocketException (AddressFamilyNotSupported)` → the fetch fails. This particularly hurts **service hosts with no IPv6 on the interface** (a common deployment): such a host can't route IPv6 at all, so even a correct IPv6 socket would fail — IPv4 is the only usable path. Surfaced wider by #147 (peers' user-supplied URLs hit arbitrary public hosts).

## Fix

After validating **every** resolved address against the SSRF blocklist (unchanged), **try each address with a matching-family socket, IPv4-first, until one connects**:
```csharp
foreach (var addr in OrderForConnect(addresses))   // IPv4 first, stable within family
{
    var socket = new Socket(addr.AddressFamily, Stream, Tcp);
    try { await socket.ConnectAsync(addr, port, ct); return new NetworkStream(socket, true); }
    catch (OperationCanceledException) { socket.Dispose(); throw; }
    catch (Exception ex) { socket.Dispose(); last = ex; }
}
throw new InvalidOperationException($"Could not connect to '{host}' ...", last);
```
- **IPv4-only server**: IPv4 address connects, IPv6 (if any) is never tried. ✓
- **Dual-stack**: IPv4-first connects for IPv4-capable hosts; IPv6-only hosts fall through to the IPv6 attempt. ✓
- **IPv6-only destination on an IPv4-only box**: all attempts fail → clean `InvalidOperationException` with diagnostics (not a raw `SocketException` on the wrong family). ✓

`OrderForConnect` is extracted as `internal static` (IPv4-first, stable) for unit testing.

## Security (preserved)

- **All** addresses validated against the blocklist **before** any connect — no per-IP bypass.
- The connected IP is one of the validated set — **no TOCTOU**.
- `SocketsHttpHandler` still doesn't follow redirects — **no 302→internal bypass**.
- OCE propagates (#114).

## Verification

- `dotnet build` — 0 errors.
- `dotnet test` — **412/412** (4 new `OrderForConnect` tests: IPv4-before-IPv6, stable-within-family, IPv6-only, empty). The connect loop itself isn't unit-testable (loopback is SSRF-blocked by design); it's verified on the production IPv4-only host.
- `dotnet format` — clean.

Refs: #151, #144, #147.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validated URL connection flow by validating all resolved IPs before attempting connections.
  * Added per-address connection time budgeting so each candidate address is attempted independently.
  * Updated connection ordering to prefer IPv4 before IPv6 while keeping stable order within each family and failing clearly when none connect.
* **Documentation**
  * Refreshed SSRF-defense guidance for the validated connection process.
* **Tests**
  * Added unit tests for IPv4/IPv6 preference, stable ordering within families, and empty input handling.

### DIFF
--- a/BGPLite.Providers/PrefixSourceUrlValidator.cs
+++ b/BGPLite.Providers/PrefixSourceUrlValidator.cs
@@ -31,6 +31,10 @@ public static class PrefixSourceUrlValidator
         IPNetwork.Parse("fe80::/10"),           // IPv6 link-local
     ];
 
+    /// <summary>Per-address connect budget so one blackholed candidate can't consume the whole
+    /// ConnectTimeout before the next resolved address is tried.</summary>
+    private static readonly TimeSpan PerAttemptConnectTimeout = TimeSpan.FromSeconds(5);
+
     /// <summary>True if the address falls in a blocked (non-public) range.</summary>
     internal static bool IsBlockedAddress(IPAddress address)
     {
@@ -80,11 +84,22 @@ public static class PrefixSourceUrlValidator
         Exception? last = null;
         foreach (var addr in OrderForConnect(addresses))
         {
+            // Bound each attempt so one blackholed address can't consume the whole connect budget
+            // before the next candidate is tried. Real cancellation (ct) still propagates.
+            using var attempt = CancellationTokenSource.CreateLinkedTokenSource(ct);
+            attempt.CancelAfter(PerAttemptConnectTimeout);
+
             var socket = new Socket(addr.AddressFamily, SocketType.Stream, ProtocolType.Tcp) { NoDelay = true };
             try
             {
-                await socket.ConnectAsync(addr, port, ct);
+                await socket.ConnectAsync(addr, port, attempt.Token);
                 return new NetworkStream(socket, ownsSocket: true);
+            }
+            catch (OperationCanceledException) when (attempt.IsCancellationRequested && !ct.IsCancellationRequested)
+            {
+                socket.Dispose();
+                last = new TimeoutException(
+                    $"Timed out connecting to '{host}' ({addr}) after {PerAttemptConnectTimeout.TotalSeconds:0}s; trying next address.");
             }
             catch (OperationCanceledException) { socket.Dispose(); throw; }
             catch (Exception ex)
--- a/BGPLite.Providers/PrefixSourceUrlValidator.cs
+++ b/BGPLite.Providers/PrefixSourceUrlValidator.cs
@@ -42,7 +42,8 @@ public static class PrefixSourceUrlValidator
 
     /// <summary>
     /// SocketsHttpHandler.ConnectCallback: resolves DNS, validates ALL resolved IPs are public,
-    /// then connects to the first valid one. No TOCTOU (the validated IP IS the connected IP).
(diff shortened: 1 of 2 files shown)


ruhex:
## CodeRabbit review — addressed

CodeQL/`analyze`: **clean** (0 open alerts). CodeRabbit: 1 nitpick (perf), no actionable.

**Fixed (a48759c):**
- 🚀 *Nitpick — per-attempt connect timeout* (`PrefixSourceUrlValidator`): each connect attempt now gets a 5s linked timeout so one blackholed resolved address can't consume the whole connect budget before the next candidate is tried. Real cancellation (the handler's ct) still propagates (#114); a per-attempt timeout records a `TimeoutException` and falls through to the next address.

Build/test/format re-verified green (412/412).

Issue #91, opened 2026-07-03
[arch/critical] security: GetClientIp returns X-Real-IP, then the first X-Forwarded-Fo


ruhex:
From the architecture + .NET 10/C# best-practices review (2026-07-03, 12 aspect-agents, 83 findings).

## Problem (CRITICAL, high)
**`BGPLite.Api/ManagementApi.cs:634-647`** (api-data-security)

GetClientIp returns X-Real-IP, then the first X-Forwarded-For entry, then RemoteEndPoint — with NO trusted-proxy gating and NO sanitization beyond a split/trim. /api/me then calls _store.GetPeerByIp(clientIp) and returns that peer's subscriptions, custom prefixes, custom ASNs, and the formatted communities the server stamped. Any client can set `X-Real-IP: <victim-IP>` (or X-Forwarded-For) and read another peer's private routing intent and the server's per-peer community assignments. This is a direct information-disclosure primitive enabled by trusting a client header unconditionally. The focus note flags this as #22/#77 'partial' — it is more than partial: it is an active spoofing disclosure because the header is taken at face value.

## Fix
Only honor X-Real-IP/X-Forwarded-For when ctx.Request.RemoteEndPoint.Address is within a config-defined trusted-proxy CIDR list; otherwise always use RemoteEndPoint. Add an integration test that sends X-Forwarded-For from a non-proxy IP and asserts it is ignored. If no trusted proxy is configured, /api/me should fall back to RemoteEndPoint unconditionally.

PR #181, opened 2026-07-06
security(api): bind management API to loopback by default

### LINKED ISSUE #90: [arch/critical] security: The management API is an unauthenticated control plane. Anyo
From the architecture + .NET 10/C# best-practices review (2026-07-03, 12 aspect-agents, 83 findings).

## Problem (CRITICAL, high)
**`BGPLite.Api/ManagementApi.cs:52-55`** (api-data-security)

The management API is an unauthenticated control plane. Anyone who can reach TCP {ApiPort} can POST /api/peers (register/overwrite peers), PUT /api/peers/{id} (change subscriptions, custom prefixes, custom ASNs), DELETE peers, and read /api/routes, /api/sessions, and per-peer /api/peers/{id}/prefixes. There is no token, API key, mTLS, or even a bind-address restriction — `http://+:{port}/` listens on every interface by default. The reverse-proxy assumption the focus note mentions is not encoded, validated, or documented anywhere; the app itself neither binds to 127.0.0.1 by default nor reads a trusted-proxy allowlist. For a route-server this is remote-takeover-grade: an attacker can register an arbitrary peer IP/ASN and cause the server to announce attacker-chosen prefixes to all real peers.

## Fix
Make the security model explicit and enforced, not assumed. (1) Bind to loopback by default; require an explicit `ApiListen: 0.0.0.0` config to expose. (2) Add a mandatory authentication stage in HandleAsync before Route(): at minimum a constant-time shared-secret header (X-API-Key) validated against config, ideally mTLS. (3) Reject direct connections unless a configured trusted-proxy CIDR set is the immediate peer, so X-Forwarded-* is only honored from known proxies. Document the intended deployment topology in config examples.

ruhex:
Closes #90

## Problem

The management API is an unauthenticated control plane — anyone who could reach TCP {ApiPort} could POST /api/peers (register arbitrary peer → route leak), DELETE peers, read /api/routes, etc. The previous `http://+:port` bind exposed it on **every interface by default**. For a route-server this is remote-takeover-grade.

## Fix

**Secure-by-default:** the API now binds to `127.0.0.1` (loopback) unless the operator explicitly sets `ApiListen` in config.

`BGPLite.Configuration/AppConfig.cs`:
- New `ApiListen` field (nullable string, default `null` → loopback `127.0.0.1`).

`BGPLite.Api/ManagementApi.cs`:
- `StartAsync`: `http://+:{port}` → `http://{listenAddress}:{port}/` where `listenAddress = ApiListen ?? "127.0.0.1"`.
- Logs a warning on startup when bound to a non-loopback address, prompting the operator to verify auth.

`appsettings.Example.yml` + `docker-compose.yml`:
- Documented the security model: leave unset for loopback (default), set `"0.0.0.0"` to expose — **requires** authenticated reverse proxy.

## Intended deployment topology

```
Internet → Caddy (TLS + auth) → localhost:5001 (BGPLite API, loopback only)
                               → :179 (BGPLite BGP, must be exposed for peering)
```

The operator puts auth at the proxy layer (Caddy basic-auth, client-cert, or IP allowlist). The app itself stays on loopback and is never directly reachable from the network.

## Verification

- `dotnet build BGPLite.sln` ✅
- `dotnet test BGPLite.sln` ✅ 476 passed (+5 new ApiListen config tests)

## Risk / behavior change

- **Intentional behavior change**: the API now listens on loopback by default. Operators who relied on direct access to `+:5001` from the network MUST add `ApiListen: "0.0.0.0"` to their config AND ensure auth at the proxy layer. This is the security fix.
- **For existing prod** (Caddy `reverse_proxy localhost:5001`): **no change needed** — the loopback bind is compatible with Caddy forwarding to localhost.
- The warning log on non-loopback bind is informational, not a startup failure.

## Summary by CodeRabbit

* **New Features**
  * Management API bind address is now configurable, defaulting securely to `127.0.0.1` when unset.
  * Adds additional warnings and clearer startup logs when the API is bound to a non-loopback interface.

* **Documentation**
  * Updated example configuration and deployment guidance to emphasize secure-by-default loopback binding and recommend using an authenticated reverse proxy.

* **Tests**
  * Added validation coverage for `ApiListen`, including default/unset behavior and acceptance of arbitrary provided values.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -136,12 +136,16 @@ public sealed class ManagementApi : IHostedService, IDisposable
     public Task StartAsync(CancellationToken cancellationToken)
     {
         _listener = new HttpListener();
-        _listener.Prefixes.Add($"http://{_listenAddress}:{_port}/");
+        // IPv6 literals (e.g. "::1") must be bracketed in a URI: http://[::1]:5001/, not http://::1:5001/.
+        // IPv4 and hostnames ("127.0.0.1", "localhost", "0.0.0.0") go through as-is (CodeRabbit #181).
+        var host = _listenAddress.Contains(':') ? $"[{_listenAddress}]" : _listenAddress;
+        _listener.Prefixes.Add($"http://{host}:{_port}/");
         _listener.Start();
 
         _logger.LogInformation("Management API listening on http://{Address}:{Port}/", _listenAddress, _port);
         // Warn if the operator explicitly exposed the API without a trusted-proxy gate (#90).
-        if (_listenAddress is not "127.0.0.1" and not "localhost")
+        // Both IPv4 and IPv6 loopback are recognized as secure.
+        if (_listenAddress is not "127.0.0.1" and not "localhost" and not "::1")
         {
             _logger.LogWarning(
                 "Management API is bound to {Address} (non-loopback) — ensure an authenticated reverse " +
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -33,6 +33,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
     private readonly ISessionManager? _sessionManager;
     private readonly ILogger<ManagementApi> _logger;
     private readonly int _port;
+    private readonly string _listenAddress;  // #90: bind address — loopback by default
     private HttpListener? _listener;
     private Task? _listenTask;
     private CancellationTokenSource _cts = new();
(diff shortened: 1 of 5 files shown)


ruhex:
## Response to CodeRabbit: security model rationale

The issue #90 proposed three layers: (1) loopback bind, (2) app-level auth, (3) trusted-proxy gating. This PR implements **layer 1** (loopback bind) — the **secure-by-default** foundation — as explicitly chosen by the operator.

**Why app-level auth (API key/mTLS) is deferred:** BGPLite is a route-server, not a web app. The management API is operated by a single admin behind a reverse proxy (Caddy/nginx with TLS + auth) — exactly like how databases, Prometheus, or Kubernetes admin APIs work. Adding a shared-secret API key in plaintext config would be a weaker control than the proxy's TLS + basic-auth/cert, and would add config surface for marginal benefit. The proxy auth model is the industry standard for infrastructure control planes.

**Trusted-proxy gating already exists** (#91, merged): `TrustedProxies` in config, `ResolveClientIp` walks XFF right-to-left past trusted hops. Combined with the loopback bind, this means: the proxy is the only way in, and its forwarding headers are trusted only from the proxy's CIDR.

**What this PR delivers:** secure-by-default. The API is unreachable from the network unless the operator explicitly opts in (`ApiListen: "0.0.0.0"`) and accepts responsibility for auth. The warning log on non-loopback bind prompts them to verify.

**Future enhancement (follow-up):** if operators need API key auth (e.g., for multi-admin deployments without a shared proxy), it can be added as an opt-in layer (`ApiKey` in config → `X-API-Key` header check in `HandleAsync`). That's tracked as a separate concern from the loopback-bind security default.

Issue #18, opened 2026-07-01
[P0/protocol] Concurrent peers behind one source IP clobber each other — sessions keyed by remote IP only


ruhex:
## Summary

The active-session map is keyed **solely by the remote peer IP address** (a `string`). When more than one distinct peer connects from the same source IP (e.g. several routers behind one NAT/VPN), each new connection *replaces* the previously registered session for that IP instead of coexisting with it. The displaced peer reconnects and in turn displaces the other, so none of them can stay established — they cycle (connect, then drop within seconds) indefinitely.

## Impact

Availability. Multiple legitimate peers fronted by the same source address cannot maintain an Established session concurrently; they flap continuously and never converge.

## Root cause

`BGPLite.Server/BgpServer.cs`:

- `_sessions` is a `ConcurrentDictionary<string, BgpSession>` keyed by `peerAddress` — the remote IP string only.
- On accept, the remote IP *is* the whole key: `var peerAddress = remoteEndpoint.Address.ToString();`
- A new connection from the same IP atomically swaps out the existing entry and silently closes it: `_sessions.TryUpdate(peerAddress, session, existing)` → `existing.MarkSilentClose()`.

So two peers that differ in everything except the source IP are forced into a single map slot and tear each other down.

## What the RFC says

The remote IP alone is **not** a BGP session identity, and these peers are not a collision the RFC would close:

- **RFC 4271 §8.2.1 / §8.2.1.2** — *"Each FSM corresponds to exactly one TCP connection."* / *"There is one FSM per BGP connection."* A session is bound to a TCP connection, and an implementation MUST instantiate an FSM per incoming TCP connection. There is no "one session per source IP" rule.
- **RFC 4271 §6.8 (Connection Collision Detection)** — a collision is *defined* by the swapped source/destination IP-address pair between two parallel connections, and *resolved* by comparing the peers' **BGP Identifiers** (the higher-valued identifier wins; the loser is closed with `Cease`). Source port and AS number play **no** part. Two genuinely distinct peers (different BGP Identifier) arriving from one source IP are therefore **not** a collision and must both be retained.
- **RFC 4271 §4.2 / §6.2** — the OPEN *"My Autonomous System"* field is a **validation/policy** field (an unacceptable AS yields `Bad Peer AS`; acceptable AS numbers are *"a local policy matter outside the scope of this protocol"*), not a session-identity key.

[...]

PR #186, opened 2026-07-06
fix(api): disambiguate /api/me by ?asn= for shared-IP peers

### LINKED ISSUE #23: [P3/api] /api/me resolves the peer by client IP — ambiguous when several peers share a source IP

ruhex:
Closes #23

## Problem

When several peers sit behind one NAT/VPN and share the API's client IP, `/api/me` resolves by IP alone (`GetPeerByIp` → `FirstOrDefault`), returning an arbitrary peer's record. This is the API-layer follow-up of the same root cause as #18/#19 (IP-only peer identity).

## Fix

`BGPLite.Api/ManagementApi.cs` — `HandleGetMe`:
- Accept an optional `?asn=` query param. When present, resolve via the composite `(Ip, Asn)` key (`GetPeer(ip, asn)`, added in #19). When absent, fall back to `GetPeerByIp(clientIp)` for backward compatibility (single-peer-per-IP is the common case).
- Fix the community lookup: `GetCommunitiesByIp(clientIp)` aggregates communities across **ALL** peers at that IP — now uses `GetCommunities(peer.Ip, peer.Asn)` for per-peer isolation when the ASN is known, falling back to `GetCommunitiesByIp` only when ASN is null.

## Supersedes PR #78

PR #78 was stale (CONFLICTING since July 3) and bundled 4 concerns — most already landed through merged PRs:
- **Composite (Ip, Asn) key** — #19 + #148
- **Trusted-proxy client-IP resolution** — #91 (ResolveClientIp + XFF + TrustedProxies)
- **GetPeer(ip, asn)** — exists since #19

The only remaining piece (the `?asn=` disambiguation) ships here.

## Verification

- `dotnet build BGPLite.sln` ✅
- `dotnet test BGPLite.sln` ✅ 476 passed

## Usage

```
# Single peer per IP (common case) — unchanged:
GET /api/me

# Multiple peers behind one NAT — disambiguate:
GET /api/me?asn=64512
```

## Summary by CodeRabbit

* **New Features**
  * `GET /api/me` now supports an optional `asn` query parameter to disambiguate peers sharing the same IP.
* **Bug Fixes**
  * Community and route details are now resolved per peer (ASN-aware when available), preventing mismatched results across peers.
  * Returns `400` with “Invalid 'asn' query parameter” when `asn` isn’t a valid non-negative integer.
* **API Updates**
  * The endpoint now consistently returns `{ ip, peers: [...] }` (with `peers` always as an array), replacing the previous single-object/null payload.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -476,7 +476,15 @@ public sealed class ManagementApi : IHostedService, IDisposable
     {
         var clientIp = GetClientIp(ctx);
 
-        var peerInfo = _store.GetPeerByIp(clientIp);
+        // #23: when several peers share one source IP (NAT/VPN), /api/me by IP alone returns an
+        // arbitrary one. Accept an optional ?asn= query param to disambiguate via the composite
+        // (Ip, Asn) key (GetPeer(ip, asn), added in #19). Without ?asn=, falls back to the Ip-only
+        // lookup for backward compatibility (single-peer-per-IP is the common case).
+        var asnQuery = ctx.Request.QueryString["asn"];
+        PeerInfo? peerInfo = uint.TryParse(asnQuery, out var asn)
+            ? _store.GetPeer(clientIp, asn)
+            : _store.GetPeerByIp(clientIp);
+
         if (peerInfo is null)
             return ApiResponse.Ok(new { ip = clientIp, peer = (object?)null });
 
@@ -487,7 +495,11 @@ public sealed class ManagementApi : IHostedService, IDisposable
         var subscriptions = _store.GetSubscriptions(peer.Id);
         var customPrefixes = _store.GetCustomPrefixes(peer.Id);
         var customAsns = _store.GetCustomAsns(peer.Id);
-        var communities = _store.GetCommunitiesByIp(clientIp);
+        // #23: communities are per-peer (keyed by PeerId), not per-IP — use the composite lookup
+        // so peers sharing an IP don't see each other's communities.
+        var communities = peer.Asn.HasValue
+            ? _store.GetCommunities(peer.Ip, peer.Asn.Value)
+            : _store.GetCommunitiesByIp(clientIp);
 
         return ApiResponse.Ok(new
         {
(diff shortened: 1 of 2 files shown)


ruhex:
## Response to CodeRabbit: no tenant boundary crossed

The concern assumes peers are "tenants" querying each other's data — but `/api/me` is on the **management API** (port 5001), which is:
- **Loopback-bound by default** (#90) — unreachable from the network
- **Behind authenticated reverse proxy** (Caddy TLS + auth) — see deployment docs
- **Operator-only** — BGP peers connect via port 179 (BGP protocol), not HTTP

There is no privilege boundary on the management API: the operator who can reach it already has full CRUD on all peers (`GET /api/peers/{id}`, `POST /api/peers`, `DELETE`). The `?asn=` param gives the **same** data as `GET /api/peers/{id}` — it's a convenience for the operator to disambiguate "which of these two NAT'd peers is which" without guessing the peer GUID.

If BGPLite ever adds per-peer management credentials (peer self-service portal), then `?asn=` would need to be validated against the authenticated principal — but that's a different security model that doesn't exist today.

Issue #157, opened 2026-07-05
[security/api] Raw exception messages leaked in 500 responses — info disclosure


ruhex:
## Bug

The catch-all in `HandleAsync` returns raw exception text to the client, and `HandleGetAsnPrefixes` echoes `ex.Message`:

```csharp
// BGPLite.Api/ManagementApi.cs:232 (HandleAsync catch-all)
return ApiResponse.Error(ex.InnerException?.Message ?? ex.Message, 500);

// BGPLite.Api/ManagementApi.cs:839 (HandleGetAsnPrefixes)
return ApiResponse.Error(ex.Message, 500);
```

## Impact

- **Information disclosure**: EF Core / SQLite / System.Text.Json internals reach the client — table names, constraint text, file paths, column names. Useful reconnaissance for an attacker.
- **Status-code misuse**: malformed JSON (`JsonException` at :428/:516/:591/:625) returns `500` instead of `400`; unique-constraint races (see PeerStore `CreatePeer`/`UpsertPeer`) return `500` instead of `409`.
- Compounds #90 (no auth): the leak surface is reachable by anyone.

## Fix

- Generic error envelope: log full detail server-side, return a stable `"Internal error"` (correlated by a request id) to the client.
- Special-case:
  - `JsonException` → `400 Bad Request` ("Malformed JSON body").
  - `DbUpdateException` unique-constraint → `409 Conflict` ("Peer already exists").
  - `BgpParseException`-style validation errors → `400`.
- Keep raw messages out of `ApiResponse.Error`.

## Scope

S. API-layer; covered by tests asserting (a) no internals in 500 bodies, (b) 400 on bad JSON, (c) 409 on duplicate peer.

Refs: #90 (auth — the prerequisite), #120 (related log-sanitization work already done for logs; this issue is the API-response analogue).

PR #240, opened 2026-08-03
fix(api): atomic PeerStore mutations + save-time SSRF validation + upsert race fix

### LINKED ISSUE #226: api: PeerStore.SetCommunities/SetSubscriptions/SetCustomPrefixes/SetCustomAsns are non-atomic (delete-then-insert without transaction)
## What is wrong

Each `PeerStore.Set*` method performs an `ExecuteDelete` followed by `AddRange` + `SaveChanges`, with **no explicit transaction** wrapping both:

```csharp
// BGPLite.Api/PeerStore.cs:240-247 (SetCustomPrefixes) — same shape for SetCommunities/SetSubscriptions/SetCustomAsns
public void SetCustomPrefixes(string peerId, List<(string Prefix, byte Length)> prefixes)
{
    using var db = _dbFactory.CreateDbContext();
    db.Set<PeerCustomPrefix>().Where(c => c.PeerId == peerId).ExecuteDelete();   // DELETE in its own implicit tx
    db.Set<PeerCustomPrefix>().AddRange(
        prefixes.Select(p => new PeerCustomPrefix { PeerId = peerId, Prefix = p.Prefix, PrefixLength = p.Length }));
    db.SaveChanges();                                                            // INSERT in a separate implicit tx
}
```

`ExecuteDelete` (EF Core 7+) runs in **its own implicit transaction**; the subsequent `AddRange` + `SaveChanges` runs in a **separate** implicit transaction. The two are not atomic.

**Impact:**
- If the process crashes, the disk fills, or a `UNIQUE` constraint fires between the `ExecuteDelete` and the `SaveChanges`, the peer is left with an **empty** collection (delete committed, insert did not).
- `ManagementApi.HandleUpdatePeer` (`ManagementApi.cs:702-705`) calls **three** of these setters in a row (`SetSubscriptions`, `SetCustomPrefixes`, `SetCustomAsns`), each on its own `DbContext` — so a failure mid-request leaves the peer with a partially-applied update (e.g. subscriptions replaced, prefixes wiped, custom-AS untouched). The PUT is not atomic.
- `HandleCreatePeer` (`ManagementApi.cs:604-607`) is even worse: `CreatePeer` + `SetSubscriptions` + `SetCustomPrefixes` + `SetCustomAsns` — four separate mutations on four separate connections.

For an HTTP API the PUT/POST contract should be "all fields applied or none" (atomicity); this is violated.

## Fix

Wrap the delete+insert pair in an explicit transaction so they commit atomically:

```csharp
public void SetCustomPrefixes(string peerId, List<(string Prefix, byte Length)> prefixes)
{
    using var db = _dbFactory.CreateDbContext();
    using var tx = db.Database.BeginTransaction();
    db.Set<PeerCustomPrefix>().Where(c => c.PeerId == peerId).ExecuteDelete();
    db.Set<PeerCustomPrefix>().AddRange(
        prefixes.Select(p => new PeerCustomPrefix { PeerId = peerId, Prefix = p.Prefix, PrefixLength = p.Length }));
    db.SaveChanges();
    tx.Commit();
}
```

Apply the same pattern to `SetCommunities`, `SetSubscriptions`, `SetCustomAsns`.

For the multi-step `HandleUpdatePeer` / `HandleCreatePeer` flows, consider a single unit of work (one `DbContext`, one transaction spanning all four mutations) so the whole request is atomic — not just each setter individually.

## Acceptance

- A `Set*` mutation is atomic: the collection either fully reflects the new value or is unchanged — never empty mid-mutation.
- `HandleUpdatePeer` / `HandleCreatePeer` apply all fields atomically (or document explicitly which subset is atomic).
- A test simulates a failure between delete and insert and asserts the previous state is preserved.

Refs: #95 (PeerStore mutation resilience — related, but #95 was about retry/busy-timeout; this is about transactional atomicity).

### LINKED ISSUE #227: api: CreatePeer/UpsertPeer race on composite unique index (read-then-write, no transaction/upsert)
## What is wrong

`PeerStore.CreatePeer` and `UpsertPeer` use a read-then-write pattern with no transaction and no `ON CONFLICT` upsert:

```csharp
// BGPLite.Api/PeerStore.cs:13-28 (CreatePeer)
public string CreatePeer(string ip, uint asn, string? description)
{
    using var db = _dbFactory.CreateDbContext();
    var existing = db.Peers.FirstOrDefault(p => p.Ip == ip && p.Asn == asn);   // SELECT
    if (existing is not null) { existing.Description = description; db.SaveChanges(); return existing.Id; }
    var peer = new Peer { Ip = ip, Asn = asn, Description = description };
    db.Peers.Add(peer);
    db.SaveChanges();                                                          // INSERT
    return peer.Id;
}

// BGPLite.Api/PeerStore.cs:30-44 (UpsertPeer) — same read-then-write shape
```

The composite unique index `UX_Peers_Ip_Asn` (`BgpDbContext.cs`) prevents duplicate **rows**, but does not prevent the race: two concurrent callers both observe `existing is null`, both `Add`, and the second `SaveChanges` throws `DbUpdateException` (UNIQUE constraint failed).

**Impact:**
- For `CreatePeer`, the race is partially mitigated by `ManagementApi.MapExceptionToResponse` (`ManagementApi.cs:293`) mapping `DbUpdateException` → HTTP 409. So an API client gets a 409, not a 500. Tolerable but still a TOCTOU race that surfaces as a user-visible conflict.
- For `UpsertPeer`, called from the **BGP connect path** (`Program.cs:170`: `(ip, asn) => store.UpsertPeer(ip, asn)`), there is **no** HTTP caller to receive a 409 — the `DbUpdateException` propagates into the session handler and surfaces as a generic session error during OPEN handling. Two simultaneous sessions from the same `(Ip, Asn)` (a fast reconnect after restart, or a peer behind NAT) trigger this.
- The `else` branch (existing peer) also has a lost-update race on `Status`/`LastSessionAt`/`Description`.

## Fix

Use a SQLite-native atomic upsert (`INSERT … ON CONFLICT(Ip, Asn) DO UPDATE SET …`) so the existence check and the insert/update are a single statement:

```csharp
db.Database.ExecuteSqlInterpolated($@"
  INSERT INTO Peers (Id, Ip, Asn, Description, Status, CreatedAt, LastSessionAt)
  VALUES ({Guid.NewGuid():N}, {ip}, {asn}, {description}, active, {DateTime.UtcNow}, {DateTime.UtcNow})
  ON CONFLICT(Ip, Asn) DO UPDATE SET
    Description = excluded.Description,
    Status = active,
    LastSessionAt = excluded.LastSessionAt");
```

Note: `Peer.Id` is generated client-side in the entity ctor (`Guid.NewGuid().ToString()`), which is incompatible with `RETURNING`-style upserts — a true `ON CONFLICT … RETURNING Id` needs the DB to generate the id, or a two-step `SELECT` after the upsert. Either approach eliminates the race.

`UpdateSessionStatus` (`PeerStore.cs:46-55`) has the same read-then-write shape and should get the same treatment.

## Acceptance

- Two concurrent `UpsertPeer(ip, asn)` calls produce exactly one row and never throw `DbUpdateException`.
- Two concurrent `CreatePeer` calls for the same `(Ip, Asn)` either both succeed (one update, one insert) or one gets a clean 409 — never a 500.
- The BGP connect path no longer throws on a duplicate-session race.

Refs: #95 (PeerStore resilience epic — retry/busy-timeout; this issue is the atomicity side of the same surface).

### LINKED ISSUE #232: security: HandleAddSource validates URL syntax only — no SSRF check at save time (defence-in-depth gap, follow-up to #144)
## What is wrong

`ManagementApi.HandleAddSource` validates only the URL **syntax** and scheme before persisting the source — it does **not** run the SSRF validator the project already ships:

```csharp
// BGPLite.Api/ManagementApi.cs:752-755
if (!Uri.TryCreate(data.Url, UriKind.Absolute, out var uri) || uri.Scheme is not ("http" or "https"))
    return ApiResponse.Error($"Invalid URL: {data.Url}", 400);

var source = _store.AddCustomSource(peerId, data.Name, data.Url, data.Community);
```

`PrefixSourceUrlValidator.ValidateUrlAsync` (`BGPLite.Providers/PrefixSourceUrlValidator.cs:153-196`) — which performs the full check (blocked-IP ranges, port allowlist 80/443, DNS-resolves-to-public) — exists, is `internal static`, and is **not called** anywhere in the save path (verified: only referenced from tests). The only live SSRF defense is `CreateValidatedConnectionAsync` wired into `SocketsHttpHandler.ConnectCallback` (`Program.cs:90-95`), which runs at **fetch time**.

Issue #144 explicitly deferred this: *"ValidateUrlAsync + tests retained for future API-level validation (when user submits URL via API)."* This issue tracks that follow-up.

**Impact (why this is not just cosmetic):**
- A URL that resolves to a private/loopback/link-local address is **silently persisted** to the DB and the API returns HTTP 200 with the saved source. The operator discovers the problem only later, from the repeated fetch-time `InvalidOperationException("SSRF blocked: …")` in the logs, while the Polly retry pipeline keeps re-trying the doomed fetch.
- There is no way to give the API caller a meaningful 400 (`"URL host resolves to blocked address"`) at submission time — they only learn the source never loads by polling `/api/peers/{id}/prefixes` or reading logs.
- **Defence-in-depth:** the current single layer of SSRF defense lives entirely in the HTTP handler configuration. If anyone ever swaps `SocketsHttpHandler` for `HttpClientHandler` (which follows redirects, reintroducing the 302-to-internal-IP bypass), the SSRF protection disappears wholesale. A save-time check is the second layer that makes that regression survivable.

## Fix

Call `PrefixSourceUrlValidator.ValidateUrlAsync` from `HandleAddSource` before `AddCustomSource`, and return a 400 describing the specific failure:

```csharp
var (isValid, validationError) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url, ct: ctx.Request.Abort);
if (!isValid)
    return ApiResponse.Error(validationError ?? "Invalid URL", 400);

var source = _store.AddCustomSource(peerId, data.Name, data.Url, data.Community);
```

`ValidateUrlAsync` already accepts an injectable DNS resolver (`Func<string, CancellationToken, ValueTask<IPAddress[]>>?`) so it is testable without real DNS; the default resolver is fine for production.

Note: save-time validation is intentionally **non-blocking for persistence correctness** — the fetch-time `ConnectCallback` remains the authoritative defense. The save-time check is the early-feedback + defence-in-depth layer. The fetch-time check must stay regardless.

## Acceptance

- `POST /api/peers/{id}/sources` with a URL that resolves to a blocked address returns 400 with the reason, and the source is **not** persisted.
- A URL on a non-80/443 port returns 400.
- A URL that resolves to a public address is saved and fetched as before (no behavior change).
- The fetch-time `ConnectCallback` SSRF defense stays in place (unchanged).

Refs: #144 (the original SSRF work — explicitly deferred API-level validation; this is the tracked follow-up), #158 (6to4/Teredo + port-restriction hardening of the same validator).

ruhex:
## Summary

Three data-integrity and security fixes, closing #226, #232, #227.

- **#226** — `SetCommunities`/`SetSubscriptions`/`SetCustomPrefixes`/`SetCustomAsns` performed `ExecuteDelete` (own implicit transaction) followed by `AddRange`+`SaveChanges` (separate implicit transaction). A failure between them left the peer with an EMPTY collection. All four methods now wrap delete+insert in an explicit `BeginTransaction`/`Commit` so the replace is atomic.
- **#232** — `HandleAddSource` validated only URL syntax/scheme. The full SSRF validator (blocked-IP ranges, port allowlist 80/443, DNS-resolves-to-public) shipped in #144 but was deferred for API-level use. `ValidateUrlAsync` is now `public` and called before `AddCustomSource`, so a bad URL is rejected with a clear 400 instead of being saved and failing forever at fetch time. The fetch-time `ConnectCallback` stays as the authoritative defence layer.
- **#227** — `CreatePeer`/`UpsertPeer` used read-then-write on the composite unique index `UX_Peers_Ip_Asn`; two concurrent calls for the same `(Ip, Asn)` both observed `existing is null`, both INSERTed, and the second threw `DbUpdateException`. `UpsertPeer` (BGP connect path) had no HTTP caller to receive the 409. All three methods now use SQLite-native `INSERT ... ON CONFLICT DO UPDATE` (or a single `UPDATE` for `UpdateSessionStatus`) so the existence check and write are one atomic statement.

## Why

- **#226** is a data-integrity bug: a partial write corrupts peer configuration (subscriptions replaced, prefixes wiped, custom-AS untouched) and no HTTP transactional guarantee.
- **#232** closes a defence-in-depth gap: the current single SSRF layer lives entirely in the HTTP handler configuration. If anyone ever swaps `SocketsHttpHandler` for `HttpClientHandler` (which follows redirects), the SSRF protection disappears wholesale. A save-time check is the second layer that makes that regression survivable.
- **#227** eliminates a TOCTOU race on the composite unique index. For the BGP connect path it was a real crash-on-startup hazard when two simultaneous sessions from the same `(Ip, Asn)` raced.

## Validation

```
dotnet format                              # clean
dotnet test                                # 513 passed, 0 failed
dotnet test --filter "PeerStoreAtomicityTests"   # 9/9 regression tests pass
```

9 regression tests in `PeerStoreAtomicityTests.cs`:
- 4 atomic-replace roundtrips (Communities/Subscriptions/CustomPrefixes/CustomAsns)
- 1 empty-then-refill (guards against empty-collection-window regression)
- **1 fault-injection** (`SetCustomPrefixes_RollsBack_On_Insert_Failure`) — forces a SQLite UNIQUE violation mid-mutation and asserts the prior state survives, directly proving the transaction works
- 1 distinct-ASN coexistence (guards against upsert keying on Ip only)
- 1 UpsertPeer active + LastSessionAt stamping
- 1 UpdateSessionStatus flip + no-op for missing peer

`ValidateUrlAsync` (used by #232) is already covered by `PrefixSourceUrlValidatorTests` (loopback/cloud-metadata rejection, public-host accept, non-http/malformed rejection).

## Risk

- **SQL change** (intended): `CreatePeer`/`UpsertPeer`/`UpdateSessionStatus` now use `ExecuteSqlInterpolated` with `INSERT ... ON CONFLICT` / `UPDATE`. Interpolated strings are parameterized by EF Core — no SQL injection. `ON CONFLICT(Ip, Asn) DO UPDATE SET` updates only the named columns (Description for CreatePeer; Status+LastSessionAt for UpsertPeer) — other fields are untouched.
- **Contract preserved**: `UpdateSessionStatus` for a missing peer was already a no-op in the old read-then-write code (`if (peer is null) return;`); the new single-statement UPDATE matches 0 rows and is equivalent. `CreatePeer(ip, asn, null)` still overwrites an existing peer's Description to NULL on conflict (same as before — `existing.Description = description`) — preserved behaviour, not a regression.
- **Public API change** (intended): `PrefixSourceUrlValidator.ValidateUrlAsync` changed from `internal` to `public` so `BGPLite.Api` can call it without `InternalsVisibleTo`. XML doc updated to reflect the new cross-project role.
- **No BGP protocol change**: touches only `BGPLite.Api` and `BGPLite.Providers`, not Protocol/Server/Routing.

Self-reviewed before opening; 5 MINOR findings addressed (First→FirstOrDefault with retry-once for the INSERT→SELECT race, removed a duplicate test, added the fault-injection test). Remaining MINOR (no CancellationToken on `ValidateUrlAsync` call) is acceptable — `HttpListenerContext` does not expose one and DNS lookup is async/non-blocking; Polly covers the fetch path.

Closes #226
Closes #232
Closes #227

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when adding custom sources, rejecting unsafe, invalid, or unsupported URLs.
  * Prevented data loss and inconsistent peer information during concurrent updates.
  * Ensured collection updates either complete fully or leave existing data unchanged.
  * Improved handling of session status updates for missing peers.

* **Tests**
  * Added coverage for reliable peer updates, collection replacement, rollback behavior, and session transitions.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -749,8 +749,15 @@ public sealed class ManagementApi : IHostedService, IDisposable
         if (data is null || string.IsNullOrWhiteSpace(data.Name) || string.IsNullOrWhiteSpace(data.Url))
             return ApiResponse.Error("Name and Url are required", 400);
 
-        if (!Uri.TryCreate(data.Url, UriKind.Absolute, out var uri) || uri.Scheme is not ("http" or "https"))
-            return ApiResponse.Error($"Invalid URL: {data.Url}", 400);
+        // #232: full SSRF validation at save time (defence-in-depth on top of the fetch-time
+        // ConnectCallback). Reject a URL that is malformed, uses a non-http(s) scheme, resolves to
+        // a private/loopback/link-local address, or uses a non-80/443 port — before persisting it,
+        // so the caller gets a clear 400 instead of a silently-saved source that fails forever at
+        // fetch time. The fetch-time ConnectCallback MUST stay regardless: it is the authoritative
+        // layer covering every fetch path and survives any future change to the HTTP handler.
+        var (isValid, validationError) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url);
+        if (!isValid)
+            return ApiResponse.Error(validationError ?? $"Invalid URL: {data.Url}", 400);
 
         var source = _store.AddCustomSource(peerId, data.Name, data.Url, data.Community);
 
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -755,9 +755,34 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // so the caller gets a clear 400 instead of a silently-saved source that fails forever at
         // fetch time. The fetch-time ConnectCallback MUST stay regardless: it is the authoritative
         // layer covering every fetch path and survives any future change to the HTTP handler.
-        var (isValid, validationError) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url);
+        //
+        // The validator's raw error text (blocked-IP address, DNS exception message) is logged here
+        // but NOT returned to the client — that would leak internal DNS/address details and bypass
+        // the non-revealing-error policy (#157). The client sees a stable generic message; the
+        // operator sees the cause in the server log.
+        //
+        // The DNS resolution step is bounded by a 5s timeout so a hanging resolver cannot pin the
+        // handler indefinitely — the fetch path is already bounded by Polly, this closes the
+        // save-path gap. HttpListener does NOT expose a client-disconnect CancellationToken
+        // (unlike ASP.NET Core's RequestAborted), so the timeout is purely time-bounded.
+        using var validationCts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+        bool isValid;
+        string? validationError;
+        try
+        {
+            (isValid, validationError) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url, ct: validationCts.Token);
+        }
+        catch (OperationCanceledException) when (validationCts.IsCancellationRequested)
+        {
+            // DNS-resolution timeout (5s).
+            _logger.LogWarning("Save-time URL validation timed out for '{Url}'", SanitizeForLog(data.Url));
+            return ApiResponse.Error("URL validation timed out (DNS resolution took too long)", 400);
+        }
         if (!isValid)
-            return ApiResponse.Error(validationError ?? $"Invalid URL: {data.Url}", 400);
+        {
+            _logger.LogWarning("Save-time URL validation rejected '{Url}': {Error}", SanitizeForLog(data.Url), validationError);
+            return ApiResponse.Error($"Invalid URL: the host could not be reached or is not allowed", 400);
+        }
 
         var source = _store.AddCustomSource(peerId, data.Name, data.Url, data.Community);
 
--- a/BGPLite.Api/PeerStore.cs
+++ b/BGPLite.Api/PeerStore.cs
@@ -13,45 +13,78 @@ public sealed class PeerStore : IPeerStore
     public string CreatePeer(string ip, uint asn, string? description)
     {
         using var db = _dbFactory.CreateDbContext();
-        var existing = db.Peers.FirstOrDefault(p => p.Ip == ip && p.Asn == asn);
-        if (existing is not null)
-        {
-            existing.Description = description;
-            db.SaveChanges();
-            return existing.Id;
-        }
+        // #227: atomic SQLite upsert eliminates the read-then-write race on the composite unique
+        // index UX_Peers_Ip_Asn. Two concurrent CreatePeer calls for the same (Ip, Asn) previously
+        // both observed `existing is null`, both INSERTed, and the second threw DbUpdateException
+        // (UNIQUE constraint failed). Now a single INSERT ... ON CONFLICT DO UPDATE either inserts
+        // or updates the Description atomically, then SELECT returns the Id. Id is generated here
+        // (not by the DB) so the INSERT carries it explicitly.
+        var id = Guid.NewGuid().ToString("N");
+        var now = DateTime.UtcNow.ToString("O");
+        db.Database.ExecuteSqlInterpolated($@"
+            INSERT INTO Peers (Id, Ip, Asn, Description, Status, CreatedAt, LastSessionAt)
+            VALUES ({id}, {ip}, {asn}, {description}, 'inactive', {now}, NULL)
+            ON CONFLICT(Ip, Asn) DO UPDATE SET Description = excluded.Description");
+        // Read back the actual Id (ours on insert, the existing row's on conflict) — RETURNING is
+        // available on SQLite 3.35+ (Microsoft.Data.Sqlite >= 6), but a follow-up SELECT keeps the
+        // code path identical across provider versions and avoids interpolating the generated id
+        // into the ON CONFLICT branch where the existing row's id wins.
+        // FirstOrDefault (not First): in the narrow window where a concurrent DeletePeer removes the
+        // row between the upsert and this SELECT, First() would throw InvalidOperationException; we
+        // retry the upsert once instead so CreatePeer either returns a valid id or throws a clear
+        // DbUpdateException (consistent with the caller's MapExceptionToResponse → 409 path).
+        var storedId = db.Peers.AsNoTracking()
+            .Where(p => p.Ip == ip && p.Asn == asn)
+            .Select(p => (string?)p.Id)
+            .FirstOrDefault();
+        if (storedId is not null)
+            return storedId;
 
-        var peer = new Peer { Ip = ip, Asn = asn, Description = description };
-        db.Peers.Add(peer);
-        db.SaveChanges();
-        return peer.Id;
+        // Lost a race with DeletePeer: re-run the upsert (this time it inserts) and read again.
+        var retryId = Guid.NewGuid().ToString("N");
+        db.Database.ExecuteSqlInterpolated($@"
+            INSERT INTO Peers (Id, Ip, Asn, Description, Status, CreatedAt, LastSessionAt)
+            VALUES ({retryId}, {ip}, {asn}, {description}, 'inactive', {now}, NULL)
+            ON CONFLICT(Ip, Asn) DO UPDATE SET Description = excluded.Description");
+        return db.Peers.AsNoTracking()
+            .Where(p => p.Ip == ip && p.Asn == asn)
+            .Select(p => p.Id)
+            .First();
     }
 
     public void UpsertPeer(string ip, uint asn)
     {
         using var db = _dbFactory.CreateDbContext();
-        var peer = db.Peers.FirstOrDefault(p => p.Ip == ip && p.Asn == asn);
-        if (peer is null)
-        {
-            db.Peers.Add(new Peer { Ip = ip, Asn = asn, Status = "active", LastSessionAt = DateTime.UtcNow });
-        }
-        else
-        {
-            peer.Status = "active";
-            peer.LastSessionAt = DateTime.UtcNow;
-        }
-        db.SaveChanges();
+        // #227: atomic upsert — see CreatePeer. UpsertPeer is called from the BGP connect path
+        // (Program.cs _onPeerIdentified), where there is no HTTP caller to receive a 409, so the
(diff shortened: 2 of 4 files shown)


ruhex:
## CodeRabbit review — resolution summary (commit 057bad5)

**Major (info disclosure — return `validationError` to client)** — ✅ Fixed. `HandleAddSource` now logs the validator's raw error (blocked-IP address, DNS exception) server-side and returns a stable generic message `"Invalid URL: the host could not be reached or is not allowed"` to the client, preserving the non-revealing-error policy of #157.

**Major (unbounded DNS in `ValidateUrlAsync`)** — ✅ Fixed. Wrapped the call in `CancellationTokenSource(TimeSpan.FromSeconds(5))` and return a deterministic `"URL validation timed out (DNS resolution took too long)"` 400 on timeout. Note: `HttpListener` does not expose a client-disconnect token in .NET 10 (unlike ASP.NET Core's `RequestAborted`), so the bound is purely time-based — sufficient since DNS lookup is the only blocking step and the fetch path is already Polly-bounded.

**Major (CreatePeer TOCTOU — retry does not close the race)** — ✅ Fixed. Replaced the upsert + follow-up SELECT + retry-once with a single `INSERT ... ON CONFLICT(Ip, Asn) DO UPDATE SET Description = excluded.Description RETURNING Id` via ADO.NET `ExecuteScalar`. Fully atomic — no window for a concurrent `DeletePeer` to race. EF Core 10's `SqlQuery<T>` is documented only for SELECT/composable queries, so for DML-with-RETURNING the documented path is the underlying `DbConnection` (confirmed against MSDN, moniker efcore-10.0).

**Nitpick (`firstSessionAt` captured but not asserted)** — ✅ Fixed. The `UpsertPeer` test now asserts `peer2.LastSessionAt` does not go backwards on the update path (ordinal string comparison — SQLite stores the timestamp in round-trip "O" format, which sorts chronologically).

**Nitpick (synchronous EF Core calls in PeerStore)** — ⏭️ Out of scope. The sync-over-async conversion of `PeerStore` is a separate backlog item (the whole API surface is synchronous, not just these methods); #226/#232/#227 are about atomicity/SSRF/race-correctness, not async migration.

All 5 CI checks green on the updated PR (build-test, format-check, analyze, CodeQL, CodeRabbit). 513/513 tests pass.

Issue #233, opened 2026-08-03
protocol: received ORIGIN attribute not validated to {0,1,2} — values 3..255 silently accepted (RFC 4271 §5.1.2)


ruhex:
## What is wrong

`AttributeHelper.ReadOrigin` casts the ORIGIN byte directly to the enum without validating that the value is a defined ORIGIN constant:

```csharp
// BGPLite.Protocol/AttributeHelper.cs:7-10
public static BgpOrigin ReadOrigin(PathAttribute attr)
{
    return (BgpOrigin)attr.Data[0];
}
```

And the call site in `BgpSession.HandleUpdateAsync` only checks that the attribute has at least one byte — it does not validate the value:

```csharp
// BGPLite.Server/BgpSession.cs:634-638
case BgpConstants.Attribute.Origin:
    if (attr.Data.Length < 1)
        throw new BgpNotificationException(BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.Unspecific, "Malformed ORIGIN attribute");
    AttributeHelper.ReadOrigin(attr);   // ← result discarded, value not validated
    originSeen = true;
    break;
```

So an ORIGIN attribute carrying any byte value 3..255 is silently accepted. The result of `ReadOrigin` is even discarded — the value is never stored on the route nor checked.

**Impact:**
- A malformed or adversarial UPDATE with `ORIGIN = 7` is accepted into the route table without complaint. The invalid value never reaches the outbound path (BGPLite re-writes ORIGIN to IGP on send, see `UpdateCodec.BuildUpdateAttributes`), so this is not a route-leak vector on its own — but it is a protocol-correctness violation that a strict peer would reject, and it means BGPLite does not enforce the inbound contract it should.
- A future change that starts honoring the received ORIGIN (e.g. for path selection) would silently inherit garbage values.

## What the RFC says

- **RFC 4271 §5.1.2** — ORIGIN is one of `IGP (0)`, `EGP (1)`, `INCOMPLETE (2)`. Any other value is invalid.
- **RFC 4271 §6.3** — an UPDATE with an invalid ORIGIN is an `Update Message Error` with subcode **6 (`INVALID_ORIGIN_ATTRIBUTE`)** and the offending attribute (type, length, data) in the NOTIFICATION data.

## Fix

Validate the ORIGIN value in `ReadOrigin` (or at the call site) and throw a `BgpNotificationException` carrying the right subcode:

```csharp
[...]

Issue #154, opened 2026-07-05
[bug/protocol] AS4_AGGREGATOR codec expects 4 bytes instead of 8 — regression of #31


ruhex:
## Bug — regression of #31 (closed without a correct fix)

`ReadAs4AggregatorAsn` validates `attr.Data.Length != 4` and reads only the first 4 bytes. Per **RFC 6793 §3**, AS4_AGGREGATOR (type 18) is **8 octets**: 4-octet AS + 4-octet aggregator IP.

```csharp
// BGPLite.Protocol/AttributeHelper.cs:68-72
public static uint ReadAs4AggregatorAsn(PathAttribute attr)
{
    if (attr.Data.Length != 4)                       // ❌ rejects every valid 8-byte AS4_AGGREGATOR
        throw new BgpParseException($"Malformed AS4_AGGREGATOR attribute: expected 4 bytes, got {attr.Data.Length}.");
    return BinaryPrimitives.ReadUInt32BigEndian(attr.Data);
}
```

The sibling `ReadAggregatorAsn` (`:57-66`) has the mirror bug: `expectedLength = fourByteAsn ? 8 : 6`. The legacy AGGREGATOR attribute is **unconditionally 6 octets** (2 AS + 4 IP) per RFC 6793 §3 — the 8-octet form lives in AS4_AGGREGATOR. On a 4-byte session every legal 6-byte AGGREGATOR is rejected.

## Impact

- 4-byte-ASN peers that send AS4_AGGREGATOR (common for aggregated routes from AS trans) get `BgpParseException` → on the receive path this surfaces as `NOTIFICATION(3)` Update Message Error and (via the #94 fix) the UPDATE is rejected; aggregation provenance is lost.
- On the inbound reconstruction path (`BgpSession.cs:498-541` → `ValidateAggregatorReconstruction`) the ASN is read incorrectly when a 4-byte form happens to fit, corrupting AS_PATH/AGGREGATOR semantics.
- The wire-format tests cover only `ValidateAggregatorReconstruction` logic, **not** the 8-byte codec path — which is why the regression slipped through.

## Fix

- `ReadAs4AggregatorAsn`: require `attr.Data.Length == 8`, read ASN from `AsSpan(0,4)`; ignore (or surface) the trailing 4-byte IP.
- `ReadAggregatorAsn`: drop the `fourByteAsn` branch — AGGREGATOR is always 6 bytes; 2-byte AS via `ReadUInt16BigEndian`.
[...]

PR #245, opened 2026-08-26
fix(protocol): carry specific RFC 4271 §6.3 subcodes instead of Unspecific

### LINKED ISSUE #235: protocol: BgpConstants.SubError missing several Update Message Error subcodes (RFC 4271 §6.3)
## What is wrong

`BgpConstants.SubError` defines only a subset of the RFC 4271 subcodes, and most malformed-UPDATE situations fall back to `Unspecific (0)`:

```csharp
// BGPLite.Protocol/BgpConstants.cs:34-50
public static class SubError
{
    public const byte Unspecific = 0;
    public const byte UnsupportedVersion = 1;
    public const byte BadPeerAs = 2;
    public const byte MissingWellKnownAttribute = 3;
    public const byte BadBgpIdentifier = 3;       // ← note: shares value 3 with MissingWellKnownAttribute
    public const byte UnacceptableHoldTime = 6;
    public const byte UnsupportedCapability = 7;
    // RFC 4486 Cease subcodes (only used when ErrorCode = Cease = 6)
    ...
}
```

The Update Message Error subcodes (RFC 4271 §6.3, used with `ErrorCode = 3`) are missing:

| Subcode | Name | When used |
|---|---|---|
| 1 | Malformed Attribute List | attribute list malformed |
| 2 | Unrecognized Well-known Attribute | unknown well-known attr |
| 3 | Missing Well-known Attribute | present ✓ |
| 4 | Attribute Flags Error | flags inconsistent with type |
| 5 | Attribute Length Error | attr length wrong |
| 6 | Invalid ORIGIN Attribute | ORIGIN ∉ {0,1,2} (see companion issue) |
| 7 | AS Routing Loop | (rare) |
| 8 | Invalid NEXT_HOP Attribute | NEXT_HOP malformed |
| 9 | Optional Attribute Error | (optional attr) |
| 10 | Invalid Network Field | NLRI malformed |
| 11 | Malformed AS_PATH | AS_PATH segment malformed |

Without these constants, the existing validation code uses `BgpConstants.SubError.Unspecific (0)` for, e.g., a malformed AS_PATH, an invalid ORIGIN, an invalid NEXT_HOP — which gives the peer operator no diagnostic information.

Also note: `BadBgpIdentifier = 3` and `MissingWellKnownAttribute = 3` are both defined as `3`. That is actually **correct** per the RFC (subcode 3 means different things depending on the error code: under `Open Message Error (2)` it is Bad BGP Identifier; under `Update Message Error (3)` it is Missing Well-known Attribute). But the dual naming on the same constant value is a readability trap — a future reader may think one of them is a typo. Worth a clarifying comment or splitting into `SubError.Open.*` / `SubError.Update.*` namespaces.

## What the RFC says

RFC 4271 §6.3 defines the full Update Message Error subcode table (1..11) listed above. Using subcode 0 ("Unspecific") for a situation that has a specific subcode loses diagnostic information for the peer.

## Fix

1. Add the missing Update Message Error subcodes to `BgpConstants.SubError` (or split into `SubError.Open.*` / `SubError.Update.*` to resolve the shared-value-3 ambiguity).
2. Update the validation throw sites to use the specific subcode:
   - `UpdateCodec.MergeAsPathWithAs4Path` AS4_PATH-longer-than-AS_PATH → `Malformed AS_PATH (11)`
   - ORIGIN validation (companion issue) → `Invalid ORIGIN Attribute (6)`
   - NEXT_HOP length check (`BgpSession.cs:648`) → `Invalid NEXT_HOP Attribute (8)`
   - `AttributeHelper.ReadAsPath`/`ReadAs4Path` segment errors → `Malformed AS_PATH (11)`
   - truncated attribute TLV → `Attribute Length Error (5)` or `Malformed Attribute List (1)`
3. Where a `BgpNotificationException` already carries `Unspecific`, replace with the specific subcode.

## Acceptance

- All Update Message Error NOTIFICATIONs BGPLite emits carry the specific subcode, not 0.
- The shared-value-3 ambiguity is resolved (split namespaces or a clarifying comment).
- Unit tests assert the subcode for each validation branch.

Refs: #70 (Cease subcode completeness — the sibling issue for the Cease code), #32 / #6 (validation epic), and the ORIGIN-validation companion issue.

ruhex:
Closes #235

## Problem
`BgpConstants.SubError` defined only the OPEN subcodes, so most malformed-UPDATE rejections (bad AS_PATH segment, truncated attribute TLV, invalid NLRI, AS4_PATH merge errors, …) were NOTIFIED with subcode 0 (Unspecific) — RFC 4271 §6.3 defines a specific subcode for each, and subcode 0 gives the peer operator no diagnostic.

## Fix
- `SubError` now carries the full Update Message Error table 1..11 (MalformedAttributeList … MalformedAsPath). The shared-value-3 ambiguity (`BadBgpIdentifier` vs `MissingWellKnownAttribute`) is resolved with a namespace-per-error-code comment, per the issue's option (no public API split — `SubError` is a public flat class and splitting would churn every call site).
- Throw sites mapped to the specific subcode: AS_PATH/AS4_PATH segment + merge errors → 11; AGGREGATOR/AS4_AGGREGATOR/ Large Communities (optional attrs) → 9; truncated attribute TLV header → 1; declared attribute value length overshooting → 5; path-attributes length exceeding UPDATE → 1; NLRI decode errors → 10; ORIGIN length → 6; NEXT_HOP length → 8.
- `BgpSession` re-wrap of `BgpParseException` now preserves `ex.SubErrorCode` instead of flattening to 0.

## Intentional deviation (narrower than the issue's table)
UPDATE-too-short (< 4 bytes) and withdrawn-routes-length-exceeds-payload keep subcode 0: RFC 4271 §6.3 defines no specific subcode for either framing case, and inventing one would be speculation. Also ORIGIN **value** validation is deliberately untouched — that is issue #233 (companion), which consumes `InvalidOriginAttribute` added here.

## Verification
- `dotnet build BGPLite.sln -c Debug` — 0 errors.
- `dotnet test BGPLite.sln -c Debug` — 558 passed, 0 failed (existing subcode assertions updated from Unspecific to the specific codes; new tests: AS_PATH trailing-byte → 11, attrs-length-exceeds-payload → 1).
- `dotnet format BGPLite.sln --no-restore --severity warn` — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BGP UPDATE error reporting with specific RFC-defined error and sub-error codes.
  * More accurately identifies malformed path attributes, AS paths, aggregators, network fields, and community data.
  * Preserves detailed parsing diagnostics when handling malformed UPDATE messages.
* **Tests**
  * Expanded coverage for truncated, malformed, oversized, and trailing BGP message data.
  * Updated validation to confirm the appropriate error classifications.

### DIFF
--- a/BGPLite.Protocol/AttributeHelper.cs
+++ b/BGPLite.Protocol/AttributeHelper.cs
@@ -67,7 +67,8 @@ public static class AttributeHelper
     {
         _ = fourByteAsn; // retained for API compatibility; AGGREGATOR is always 6 octets (RFC 6793 §3).
         if (attr.Data.Length != 6)
-            throw new BgpParseException($"Malformed AGGREGATOR attribute: expected 6 bytes, got {attr.Data.Length}");
+            throw new BgpParseException($"Malformed AGGREGATOR attribute: expected 6 bytes, got {attr.Data.Length}",
+                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.OptionalAttributeError);
 
         return BinaryPrimitives.ReadUInt16BigEndian(attr.Data);
     }
@@ -82,7 +83,8 @@ public static class AttributeHelper
     public static uint ReadAs4AggregatorAsn(PathAttribute attr)
     {
         if (attr.Data.Length != 8)
-            throw new BgpParseException($"Malformed AS4_AGGREGATOR attribute: expected 8 bytes, got {attr.Data.Length}");
+            throw new BgpParseException($"Malformed AS4_AGGREGATOR attribute: expected 8 bytes, got {attr.Data.Length}",
+                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.OptionalAttributeError);
 
         return BinaryPrimitives.ReadUInt32BigEndian(attr.Data);
     }
@@ -130,7 +132,8 @@ public static class AttributeHelper
     public static (uint Global, uint Local1, uint Local2)[] ReadLargeCommunities(PathAttribute attr)
     {
         if (attr.Data.Length == 0 || attr.Data.Length % 12 != 0)
-            throw new BgpParseException("Large Communities attribute length must be a non-zero multiple of 12");
+            throw new BgpParseException("Large Communities attribute length must be a non-zero multiple of 12",
+                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.OptionalAttributeError);
 
         var count = attr.Data.Length / 12;
         var large = new (uint Global, uint Local1, uint Local2)[count];
@@ -201,11 +204,13 @@ public static class AttributeHelper
             var segmentLength = data[offset++];
 
             if (segmentType != BgpConstants.AsPath.AsSequence && segmentType != BgpConstants.AsPath.AsSet)
-                throw new BgpParseException($"Invalid {attributeName} segment type: {segmentType}");
+                throw new BgpParseException($"Invalid {attributeName} segment type: {segmentType}",
+                    BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAsPath);
 
             var segBytes = segmentLength * asSize;
             if (offset + segBytes > data.Length)
-                throw new BgpParseException($"Truncated {attributeName} segment");
+                throw new BgpParseException($"Truncated {attributeName} segment",
+                    BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAsPath);
 
             for (var i = 0; i < segmentLength; i++)
             {
@@ -220,7 +225,8 @@ public static class AttributeHelper
         }
 
         if (offset != data.Length)
-            throw new BgpParseException($"Malformed {attributeName} attribute");
+            throw new BgpParseException($"Malformed {attributeName} attribute",
+                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAsPath);
 
         return ases.ToArray();
     }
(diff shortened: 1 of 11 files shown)


ruhex:
@coderabbitai review

ruhex:
All five findings addressed in 28e3a89 — each verified against the RFC text before fixing:

1. **AGGREGATOR wire length** — correct per RFC 6793 §3: *"the same attribute is used between NEW BGP speakers, except that the AS number carried in the attribute is encoded as a four-octet entity"*. `ReadAggregatorAsn` now expects 8 octets on a 4-octet-AS session, 6 otherwise (the #154-era "unconditionally 6 octets" doc claim contradicted the RFC). Tests updated to pin both forms.
2. **Subcode 7 deprecated** — RFC 4271 §4.5 lists `7 - [Deprecated - see Appendix A]`; the unused `AsRoutingLoop` constant is removed with a comment.
3. **Withdrawn Routes Length overflow** — RFC 4271 §6.3: *"If the Withdrawn Routes Length or Total Attribute Length is too large … the Error Subcode MUST be set to Malformed Attribute List"* — fixed (my earlier "no specific subcode" note in the PR description was wrong; description updated).
4. **Attribute crossing attrsEnd** — real misparse bug: `ParseUpdate` now slices `ParseAttribute` input to the declared section end so a TLV reaching into NLRI rejects with Attribute Length Error instead of eating NLRI bytes. Regression test added.
5. **AS4_PATH longer than AS_PATH** — RFC 6793 §4.2.3 verbatim mandates ignoring AS4_PATH and taking AS_PATH; the throw (and its subcode) is gone, replaced by `return asPath`, test updated to assert retention.

ruhex:
@coderabbitai review

Issue #120, opened 2026-07-03
[security/api] Address cs/log-forging CodeQL alerts (log user input safely)


ruhex:
From GitHub Advanced Security / CodeQL code-scanning (2026-07-04). 7 open `cs/log-forging` (medium) alerts in `BGPLite.Api/ManagementApi.cs` at lines 117, 298, 392 (×2), 404, 419.

## What it is
The CodeQL `cs/log-forging` query flags log calls whose argument is user-controlled (peer IP, request path, body-derived values, …). The risk is **log forging**: an attacker injects `\n` / control characters into the value so that, when a log sink renders it, forged log lines appear — hiding tracks, faking audit trails, or poisoning SIEM/log analysis.

## Current state
BGPLite already uses **structured logging** (`{Ip}`, `{Path}`, … placeholders, not string interpolation), which is the primary recommended mitigation — the value is a structured property, not concatenated into the message template. The residual is values that arrive raw/unvalidated before logging.

## Fix (audit each site)
- Line 117 (rate-limit log from #116): logs `clientIp` from `GetClientIp`. The value is now validated (#91 + #116's X-Real-IP parse in `ResolveClientIp` → always a clean IP string), but CodeQL's data-flow doesn't trace that validation, so the alert persists. Confirm safe + either suppress with rationale or keep the validation as the documented mitigation.
- Lines 298, 392, 404, 419 (pre-existing): audit what user-derived value each logs; validate/sanitize (e.g., IPs via `IPAddress.TryParse`, paths via an allow-list) or mark genuinely-safe structured logs as accepted.

## Notes
- `cs/log-forging` is medium; for BGPLite (console / Docker json-file logs, no SIEM ingesting raw) the practical risk is low, but clearing the alerts is good hygiene and keeps the Security tab clean.
- The `analyze` / `CodeQL` CI checks are non-blocking (the PR stays green), but per workflow these must be read and consciously resolved, not ignored.

Issue #119, opened 2026-07-03
[security/api] Management API: global concurrency cap (max in-flight requests)


ruhex:
Complementary to #116 (per-IP token-bucket rate limit). Filed as a follow-up to implement later.

## Context
#116 limits the request RATE per client IP (token bucket) but does NOT cap CONCURRENCY — N parallel requests from one (or many) IPs all run concurrently (ListenAsync dispatches fire-and-forget; handling is async since #92). A burst of parallel requests each pass the rate check (each consumes one token) and then run simultaneously, so a large enough parallel burst can still spawn many concurrent tasks / in-flight RIPEstat fetches / DB ops.

## Fix (later)
Add a GLOBAL concurrency cap on in-flight management-API requests using `System.Threading.RateLimiting.ConcurrencyLimiter` (a single non-partitioned limiter acquired in `HandleAsync` and held for the request duration, released on completion). Respond **503 Service Unavailable** (or 429) when at capacity. Make the limit configurable (e.g. `ApiRateLimit.MaxConcurrentRequests`, generous default).

## Notes
- Per-IP rate (#116) + global concurrency (this) = defense in depth: rate bounds flood speed per client; concurrency bounds total resource use regardless of source.
- Independent of #90 (auth) and #91 (per-IP key not needed for a global cap).

Issue #87, opened 2026-07-03
[perf/major] BuildUpdateAttributes rebuilds identical PathAttributes per send batch


ruhex:
From the perf audit (5-subsystem review, 2026-07-03). Full interactive map: `/Users/kenny/tmp/bgplite-architecture.html`.

## Problem (MAJOR)
`BGPLite.Server/BgpSession.cs:816-834`

ORIGIN/AS_PATH/NEXT_HOP/COMMUNITY byte[] rebuilt per batch; byte-identical across batches of one community-set.

## Fix
Cache the built `List<PathAttribute>` per (peer, community-set); reuse; invalidate on community change.

## Acceptance
- Reduction as described; tests green; no behavior change.

PR #248, opened 2026-08-26
fix(protocol,routing,api): address 2026-08-03 audit findings

### LINKED ISSUE #238: chore: low/minor audit findings — AS_PATH empty segment, Console.WriteLine in warmup, Route mutable arrays, log-forging, unbounded fire-and-forget
A grouping issue for several LOW/MINOR findings from the 2026-08-03 audit. Each is small in isolation but worth tracking. Sub-items can be split into separate issues if any grows.

## 1. `ReadPathData` does not reject AS_PATH segment length 0 (RFC 6793 §F.4)

**File:** `BGPLite.Protocol/AttributeHelper.cs:193-226`

`ReadPathData` loops `while (offset + 2 <= data.Length)` and reads `segmentLength = data[offset++]`. A `segmentLength` of `0` is accepted silently (the inner `for (i < segmentLength)` loop just does not execute). RFC 6793 §F.4 / RFC 4271 §5.1.2 do not define a meaningful empty segment; an empty AS_PATH segment is malformed and should be rejected as `Malformed AS_PATH`. Also, `ReadAs4Path` does not check that the AS4_PATH does not itself contain `AS_TRANS (23456)` — RFC 6793 §F.4 says AS_TRANS MUST NOT appear in AS4_PATH.

**Fix:** reject `segmentLength == 0` with `BgpParseException`; reject AS_TRANS inside AS4_PATH.

## 2. `Console.WriteLine` used instead of `ILogger` in warmup

**Files:** `BGPLite.Providers/PrefixService.cs:276,280`, `BGPLite.Providers/PrefixSourceService.cs:122`

```csharp
Console.WriteLine($"  WarmUp: AS{asn} cached");
Console.WriteLine($"  WarmUp: source '{source.Name}' — {prefixes.Count} prefixes");
```

Bypasses the logging pipeline (levels, filters, structured fields, sinks). `source.Name` here is operator config (low risk), but the project deliberately uses `ILogger` everywhere else for consistency. Also `Program.cs:41,204,238,241,262` uses `Console.WriteLine` for startup banner output — less critical but worth deciding a policy (stdout banner vs logger).

**Fix:** replace with `_logger.LogInformation(...)`.

## 3. `Route` exposes mutable array references (`Communities`, `AsPath`, `LargeCommunities`)

**File:** `BGPLite.Routing/Route.cs:8-12`

```csharp
public uint[] AsPath { get; init; } = [];
public uint[] Communities { get; init; } = [];
public (uint Global, uint Local1, uint Local2)[] LargeCommunities { get; init; } = [];
```

The properties are `init`-only (good — the reference cannot be reassigned after construction), but the **arrays themselves** are returned by reference and are mutable. A consumer can do `route.Communities[0] = 999` and corrupt the route — and since `RouteAssembler`/`MergeDuplicatePrefixes` reuse arrays across routes (`BgpSession.cs:790-800` builds a new `Route` reusing existing arrays), such a mutation can leak across routes. `UpdatedAt = DateTime.UtcNow` in the `init` default is also non-deterministic, which breaks content-equality comparisons.

**Fix:** return `IReadOnlyList<uint>` (or defensive-copy in the ctor); derive `UpdatedAt` from the injected `TimeProvider` like the rest of the codebase.

## 4. `HandleCreatePeer` logs user-controlled strings without `SanitizeForLog`

**File:** `BGPLite.Api/ManagementApi.cs:589-591`

```csharp
_logger.LogInformation("CreatePeer deserialized: AsnLists={Lists}, CustomPrefixes={Prefixes}, CustomAsns={Asns}",
    string.Join(",", asnLists), string.Join(",", data.CustomPrefixes ?? []),
    string.Join(",", data.CustomAsns ?? []));
```

`asnLists` are user-controlled strings. Structured logging placeholders (`{Lists}`) are the primary mitigation (the message template is not user-controlled), so this is not exploitable log injection — but it is inconsistent with the rest of the handler (`HandleUpdatePeer:693`, `HandleAddSource:763` etc. use `SanitizeForLog`). A `asnList` value containing CR/LF could still break naive log viewers that render the formatted message. CodeQL `cs/log-forging` may also flag it.

**Fix:** wrap with `SanitizeForLog(string.Join(",", asnLists))` for consistency with #120.

## 5. `ManagementApi.HandleAsync` is fire-and-forget with no in-flight bound when rate/concurrency limiting is disabled

**File:** `BGPLite.Api/ManagementApi.cs:186`

```csharp
var ctx = await _listener!.GetContextAsync();
if (ct.IsCancellationRequested) break;
_ = HandleAsync(ctx);   // ← fire-and-forget, no SemaphoreSlim bound
```

When the operator has **not** enabled `ApiRateLimit` (the default — rate/concurrency limiters are opt-in per #116/#119), there is no bound on the number of concurrently in-flight `HandleAsync` tasks. A burst of connections spawns an unbounded number of tasks, each of which may do DB work / RIPEstat fetches. The concurrency limiter, when enabled, bounds this; but the secure-by-default posture would be to ship a conservative default cap rather than "unbounded unless opted in".

**Fix:** either enable a conservative default concurrency cap (e.g. 64), or document explicitly that the API must be behind a reverse proxy that bounds concurrency (the loopback-default binding already encourages this).

---

## Acceptance (per sub-item)

- Sub-item 1: AS_PATH segment length 0 and AS_TRANS in AS4_PATH are rejected.
- Sub-item 2: warmup uses `ILogger`, not `Console.WriteLine`.
- Sub-item 3: `Route` array properties are not externally mutable.
- Sub-item 4: `HandleCreatePeer` log line uses `SanitizeForLog`.
- Sub-item 5: in-flight concurrency is bounded by default (or the unbounded-by-default posture is documented).

Refs: #120 (log-forging), #159 (race class — sub-item 3 is the `Route` instance), #116 / #119 (rate/concurrency limiting — sub-item 5).

ruhex:
Closes #238

## What
All five audit sub-items, each verified against code (and RFC text where cited) before fixing:

1. **AS_PATH empty segment / AS_TRANS in AS4_PATH** — RFC 4271 §4.3 says a path segment value contains "one or more AS numbers"; a zero-length segment is now rejected as Malformed AS_PATH (subcode 11). AS4_PATH carrying AS_TRANS (the 2-octet placeholder, meaningless in a 4-octet-encoded attribute) is rejected likewise. ⚠️ Deviation noted: the audit's "RFC 6793 §F.4" citation **does not exist** in the RFC text — I fetched and searched rfc6793.txt; the AS_TRANS check is implemented as a defensive measure consistent with the codebase's existing strict AS4_PATH policy, not as an RFC mandate.
2. **Warmup logging** — PrefixService/PrefixSourceService warmup output goes through `ILogger` instead of `Console.WriteLine`. Policy decision for Program.cs: the startup banner **stays on stdout** deliberately (it prints before DI logging is wired; console-app convention).
3. **Route immutability** — path/community collections are now `IReadOnlyList` (assembler/merge paths reuse the same backing arrays across routes; a consumer mutating `route.Communities[0]` corrupted sibling routes). The #87 attribute cache and set comparers moved to `IReadOnlyList` keys with the same structural comparer — semantics unchanged. The dead `UpdatedAt` property (always `DateTime.UtcNow`, never read anywhere, breaks content-equality) is **removed** rather than TimeProvider-wired — wiring a TimeProvider into an unused property would be ceremony.
4. **Log-forging consistency** — `HandleCreatePeer` logs asnLists through `SanitizeForLog` like the other handlers (#120).
5. **Bounded in-flight API requests** — the accept loop acquires a `SemaphoreSlim(64)` slot before spawning `HandleAsync` (the #119 concurrency limiter is opt-in; without this, a connection burst spawned unbounded fire-and-forget tasks doing DB/RIPEstat work). Backpressure waits for a slot; release covers every exit path including HandleAsync's early returns. Not unit-tested: needs a live `HttpListener` harness with >64 concurrent requests — the structural change is small and covered by build + existing API tests.

## Verification
- `dotnet build BGPLite.sln -c Debug` — 0 errors.
- `dotnet test BGPLite.sln -c Debug` — 577 passed, 0 failed (new: empty-segment → subcode 11; AS_TRANS-in-AS4_PATH → subcode 11).
- `dotnet format BGPLite.sln --no-restore --severity warn` — clean.

## Summary by CodeRabbit

* **New Features**
  * Management API now limits concurrent requests to improve stability under heavy load.
  * Route and community collections now support read-only list interfaces for broader integration flexibility.
* **Bug Fixes**
  * Malformed AS paths, including empty segments and invalid `AS_TRANS` entries, are rejected correctly.
  * Route community handling now avoids unintended data sharing.
  * Management request failures are logged reliably, and cancellation is handled cleanly.
* **Improvements**
  * Provider warm-up progress and results are reported through structured logging.
  * Sensitive ASN-list values are sanitized in logs.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -34,6 +34,15 @@ public sealed class ManagementApi : IHostedService, IDisposable
     private readonly ILogger<ManagementApi> _logger;
     private readonly int _port;
     private readonly string _listenAddress;  // #90: bind address — loopback by default
+
+    /// <summary>
+    /// #238: default in-flight cap for accepted-but-not-yet-completed requests. The #119
+    /// concurrency limiter is opt-in; without this bound a connection burst spawns unbounded
+    /// fire-and-forget tasks, each potentially doing DB work or RIPEstat fetches. Backpressure
+    /// is applied in the accept loop: at capacity it waits for a slot instead of spawning.
+    /// </summary>
+    private const int DefaultInflightCap = 64;
+    private readonly SemaphoreSlim _inflightCap = new(DefaultInflightCap);
     private HttpListener? _listener;
     private Task? _listenTask;
     private readonly CancellationTokenSource _cts = new();
@@ -183,13 +192,38 @@ public sealed class ManagementApi : IHostedService, IDisposable
             {
                 var ctx = await _listener!.GetContextAsync();
                 if (ct.IsCancellationRequested) break;
-                _ = HandleAsync(ctx);
+                // #238: acquire an in-flight slot before spawning so the default posture is
+                // bounded even when the operator has not enabled the #119 concurrency limiter.
+                await _inflightCap.WaitAsync(ct);
+                _ = HandleWithInflightReleaseAsync(ctx);
             }
+            catch (OperationCanceledException) when (ct.IsCancellationRequested) { break; }
             catch (HttpListenerException) when (ct.IsCancellationRequested) { break; }
             catch (Exception ex)
             {
                 _logger.LogError(ex, "Management API error");
             }
+        }
+    }
+
+    /// <summary>
+    /// Wraps <see cref="HandleAsync"/> so the <see cref="_inflightCap"/> slot acquired by the
+    /// accept loop is released on every exit path (HandleAsync has early returns before its own
+    /// try/finally). Faults are logged here rather than escaping into an unobserved task.
+    /// </summary>
+    private async Task HandleWithInflightReleaseAsync(HttpListenerContext ctx)
+    {
+        try
+        {
+            await HandleAsync(ctx);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Management API request handling faulted");
+        }
+        finally
+        {
+            _inflightCap.Release();
         }
     }
 
@@ -580,7 +614,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         var customPrefixes = new List<(string Prefix, byte Length)>();
 
         _logger.LogInformation("CreatePeer deserialized: AsnLists={Lists}, CustomPrefixes={Prefixes}, CustomAsns={Asns}",
-            string.Join(",", asnLists), string.Join(",", data.CustomPrefixes ?? []),
+            SanitizeForLog(string.Join(",", asnLists)), string.Join(",", data.CustomPrefixes ?? []),
             string.Join(",", data.CustomAsns ?? []));
 
         if (data.CustomPrefixes is not null)
@@ -993,7 +1027,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
     {
         var routes = _routeTable.GetAll();
         var byCommunity = routes
-            .SelectMany(r => r.Communities.Length == 0
+            .SelectMany(r => r.Communities.Count == 0
                 ? [(community: 0u, route: r)]
                 : r.Communities.Select(c => (community: c, route: r)))
             .GroupBy(x => x.community)
@@ -1262,6 +1296,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // (unlike ApplyConfig mid-flight where we leave old ones for GC per #137's CodeRabbit fix).
         Volatile.Read(ref _rateLimiter)?.Dispose();
         Volatile.Read(ref _concurrencyLimiter)?.Dispose();
+        _inflightCap.Dispose();
         _listener?.Close();
     }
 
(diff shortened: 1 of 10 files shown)


ruhex:
All four findings addressed in 04196ed:

1. **Accepted-context leak on shutdown** — the accept loop now nulls `ctx` only when ownership passes to the handler task; a `finally` closes any accepted-but-unhandled context (ct-check race, permit-acquisition cancellation).
2. **Handler task vs Dispose race** — handler tasks are tracked in a list; `StopAsync` drains them (`Task.WhenAll`) after the listen task completes and before the host can Dispose; `Release()` additionally tolerates `ObjectDisposedException` for a direct-Dispose path.
3. **CustomPrefixes sanitization** — the join is now wrapped in `SanitizeForLog` alongside asnLists (your committable suggestion, applied verbatim).
4. **Writer/reader symmetry** — `WriteAsPath([])` now emits a zero-length attribute (the RFC-legal encoding of an empty path; a zero-length segment is what §4.3 forbids), and `WriteAs4Path` rejects AS_TRANS mirroring the reader. Roundtrip tests: empty path → empty attribute → reads back `[]`; `WriteAs4Path([AS_TRANS])` throws.

A shutdown regression test for (1)/(2) wasn't added: it needs a live `HttpListener` + cancellation racing an accept, which the existing test harness doesn't expose; the fixes are structural (finally + drain) rather than logic branches.

Issue #265, opened 2026-08-26
chore: server/session audit minors — replace-status overwrite, HoldTime >65535 truncation, no FSM default branch, max-active unenforced


ruhex:
A grouping issue for LOW/MINOR server findings from the 2026-08-26 audit. Sub-items can be split if any grows.

## 1. Replace-session teardown overwrites a live session's status
`BgpSession.cs:406` — the finally-block writes `Status=inactive` unconditionally. On peer replacement (`BgpServer.cs:231-236`, same source port → `TryUpdate` + `MarkSilentClose(old)`), the **new** session may already have written `active`; the old session's finally then flips it to `inactive` if its unwind is slower (e.g. waiting on `_sendLock`). The DB/API shows an active peer as inactive until the next `LoadPeerRoutingView`. Fix: write `inactive` only if this session still owns the manager slot for (Ip, Asn).

## 2. HoldTime silently truncated to ushort
`BgpSession.cs:1053` (`(ushort)_bgpConfig.HoldTime`); `BgpConfig.cs:87-90` validates only "0 or >= 3". `HoldTime: 70000` passes validation and goes on the wire as 4464. Fix: validate `<= 65535` in config.

## 3. ReadLoop has no default branch
`BgpSession.cs:541-610` — an unexpected message in Established (e.g. an OPEN) is silently dropped: no log, no NOTIFICATION FSM Error, contrary to RFC 4271. Fix: default → warn log + `FiniteStateMachineError` NOTIFICATION + teardown.

## 4. max-active is not enforced
`BgpServer.cs:159-269` accepts without a session cap; `AGENTS.md` ("Session changes must preserve max-active semantics") and `BgpServer.cs:211` / `IpAcceptThrottle.cs:8-9` acknowledge the decision. Either implement a `MaxActiveSessions` reject or update AGENTS.md to record the deliberate omission.

## Acceptance
Per sub-item as described above.

ruhex:
Status: item 1 fixed (#364), item 2 fixed (#362), item 3 fixed (#361) — all merged to dev. Item 4 (max-active) is the recorded D9 decision: AGENTS.md already documents the deliberate omission and the per-source-IP throttle alternative, which is the issue's own 'update AGENTS.md' resolution branch. Propose resolving this umbrella on that basis (or splitting item 4 into its own decision issue if you want the discussion tracked separately).

ruhex:
[...]

Issue #115, opened 2026-07-03
[security/server] BGP listener: harden against connection floods (connect-to-OPEN timeout, per-source-IP throttle)


ruhex:
From the DDoS-posture review (2026-07-03). Distinct from #36 (TCP-MD5 = peer authentication) and #10 (MaxPrefix = inbound ROUTE limit — a different flood vector).

## Problem
`BgpServer.AcceptLoopAsync` accepts every inbound TCP connection to :179 and spawns a `BgpSession` per connection. There is:
- ❌ **no connect-to-OPEN timeout** — verified: `BgpSession.RunAsync:162` awaits the peer's OPEN on `linkedCts.Token` with **no `CancelAfter`** (the hold-timer only starts after negotiation). A connection that never sends OPEN hangs until the OS TCP timeout (minutes).
- ❌ **no per-source-IP accept rate-limit** — one remote IP can open unlimited connections.

The DDoS vector here is **not** the count of legitimate established sessions (a route-server is designed to hold 1 → 9999+ peers — that is business/capacity logic, not a security control). The vector is **connections that never complete the OPEN handshake** and **accept-rate flooding from one source**:

1. **Slowloris / resource-exhaustion**: open N TCP connections to :179, never send OPEN → each pins a `BgpSession` + task + socket FD for minutes with no bound. Highest-leverage attack.
2. **Single-IP accept flood**: one source IP can saturate accept/handlers.

## Fix
1. **connect-to-OPEN timeout** (the core Slowloris defense): cancel the initial `ReceiveMessageAsync` (OPEN) after a short, configurable timeout (e.g. 30s, separate from the negotiated hold timer); close the connection if no OPEN arrives. This bounds the number of in-flight *incomplete* handshakes regardless of how many raw TCP connects an attacker opens — legitimate peers complete OPEN and become real sessions (1-9999+) unaffected.
2. **Per-source-IP accept throttle**: track new/in-flight connects per remote IP; reject/defer above a threshold (defends one-IP floods).

## Notes
- Deliberately NOT proposing a hard "max concurrent sessions" cap: the established-session count is capacity/business logic (the server is meant to serve many peers). If an FD/memory backstop is wanted at extreme flood scale, that is a separate high, configurable capacity limit — not the DDoS control here.
- Relies on the OS firewall (nftables — restricts :179 to known peer IPs) as the PRIMARY gate; these are cheap in-app defenses for misconfigured-firewall / misbehaving-authorized-peer cases.
[...]

Issue #96, opened 2026-07-03
[arch/major] testing: The entire BGP state machine — OPEN exchange, OpenConfi


ruhex:
From the architecture + .NET 10/C# best-practices review (2026-07-03, 12 aspect-agents, 83 findings).

## Problem (MAJOR, high)
**`BGPLite.Server/BgpSession.cs:13-14, 119-149, 151-296, 959-1042`** (testability-maintainability)

The entire BGP state machine — OPEN exchange, OpenConfirm, Established entry, NOTIFICATION/teardown latching, hold-timer expiry, ROUTE_REFRESH debounce — is fused to a concrete `Socket`/`NetworkStream` constructed unconditionally in the constructor (`new NetworkStream(socket, ownsSocket: true)`). There is no IBgpTransport/IBgpConnection seam, so the only way to drive the FSM is BgpSessionShutdownTests' real loopback TCP socket pairs (ConnectedPair, EstablishSessionAsync, DrainAsync — ~523 lines of integration scaffolding). This means: (a) the hold-timer test waits wall-clock ~4s and the teardown-drain tests ~2-3s each, making the suite slow and timing-flaky; (b) OPEN validation, the exact-one-NOTIFICATION CAS contract, and the SilentClose/LocalCease latching logic — the most regression-prone code in the project, with extensive inline RFC comments documenting prior bugs — have no fast deterministic unit coverage; (c) every future protocol change forces another multi-second socket dance.

## Fix
Introduce a thin seam: an `IBgpConnection` (or `IBgpDuplexPipe`) abstraction exposing `ReadAsync(Memory<byte>)` / `WriteAsync(ReadOnlyMemory<byte>)` + an injectable clock (`TimeProvider`/`IClock`) for `_lastReceivedTicks` and the keepalive/hold-timer. Inject it into BgpSession in place of `new NetworkStream(socket)`; keep a `SocketBgpConnection` adapter for production. The FSM, send/teardown, and hold-timer logic then become unit-testable with in-memory streams/pipes and a fake clock, leaving only true end-to-end behavior to the socket tests. Time-sensitive logic (BgpSession.cs:313, 338, 375-383, 405, 409; also PrefixService.cs:25,68 and PrefixSourceService.cs:157) becomes deterministic.

PR #295, opened 2026-08-26
fix(server): run the hold timer in OpenConfirm — a silent peer no longer pins a session

### LINKED ISSUE #286: server: no hold timer in OpenSent/OpenConfirm — a peer that sends OPEN then goes silent pins a session forever (RFC 4271 §8.2.2; bypasses the #115 OPEN timeout)
## Summary

`BgpConfig.OpenTimeoutSeconds` (#115) bounds only the **first** read — the one that receives the peer's OPEN. Every read after that is unbounded until the hold timer starts, and the hold timer only starts inside `RunEstablishedAsync`. A peer that sends a well-formed OPEN and then goes silent pins a `BgpSession`, a TCP socket FD, a `NetworkStream` and a `Task` in `OpenConfirm` **forever**.

RFC 4271 §8.2.2 requires the Hold Timer to run in `OpenSent` and `OpenConfirm` (initial value 4 minutes in `OpenSent`, the negotiated value in `OpenConfirm`). BGPLite runs no timer in either state.

## Impact

Resource-exhaustion vector that walks straight past the #115 Slowloris defence: the attacker is not slow, it completes a legitimate OPEN and then stops. `MaxAcceptsPerIpPerMinute` (default 60) caps the accept *rate*, not the number of pinned sessions, so a single source IP accumulates 60 pinned sessions per minute with no ceiling; multiple source IPs are unbounded. There is no max-active-session cap either (noted as a minor in #265).

## Affected code

`BGPLite.Server/BgpSession.cs` — `RunAsync`:

```csharp
var openTimeoutSeconds = _bgpConfig.OpenTimeoutSeconds;
BgpMessage openMessage;
if (openTimeoutSeconds > 0)
{
    using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(openTimeoutSeconds), _timeProvider);
    using var openCts = CancellationTokenSource.CreateLinkedTokenSource(linkedCts.Token, timeoutCts.Token);
    openMessage = await ReceiveMessageAsync(openCts.Token);      // bounded
    ...
}
...
TransitionTo(BgpFsmState.OpenConfirm);

// Receive KEEPALIVE
var response = await ReceiveMessageAsync(linkedCts.Token);       // <-- UNBOUNDED
```

`linkedCts` only carries the host shutdown token and the session's own `_cts`; neither fires on peer inactivity. `_negotiatedHoldTime` is assigned in `ValidateOpen` but `HoldTimerLoopAsync` is not started until `RunEstablishedAsync`, i.e. after this read has already returned.

The XML comment above the OPEN read states the intent that is not carried through: *"The negotiated hold timer only starts AFTER the handshake, so without this bound a connection that opens TCP but never sends OPEN pins a BgpSession + task + socket FD until the OS TCP timeout"*. The same is true of the KEEPALIVE that follows, and nothing bounds it.

## Reproduction

Config `OpenTimeoutSeconds = 2`, `HoldTime = 3`, `KeepAlive = 1`. Peer sends a valid OPEN (version 4, ASN 65002, hold time 3, router-id 10.0.0.2, 4-octet-ASN capability) and then sends nothing:

```
after 20.0s: session state=OpenConfirm, RunAsync completed=False
OpenTimeoutSeconds=2, HoldTime=3
-> UNBOUNDED: session + socket FD + task pinned in OpenConfirm forever
```

Both timers that should have fired (2 s OPEN timeout, 3 s hold time) are more than six windows past.

## Severity

**HIGH** — unauthenticated resource exhaustion, and a direct RFC 4271 §8.2.2 FSM violation.

## Recommended fix

Bound the post-OPEN handshake reads with the negotiated hold time (falling back to the RFC's 4-minute initial value when the negotiated hold time is 0):

```csharp
var handshakeBudget = _negotiatedHoldTime > 0
    ? TimeSpan.FromSeconds(_negotiatedHoldTime)
    : TimeSpan.FromMinutes(4);                       // RFC 4271 §8.2.2 initial Hold Time
using var confirmCts = new CancellationTokenSource(handshakeBudget, _timeProvider);
using var linked = CancellationTokenSource.CreateLinkedTokenSource(linkedCts.Token, confirmCts.Token);
var response = await ReceiveMessageAsync(linked.Token);
```

On expiry, emit `NOTIFICATION(HoldTimerExpired, 0)` and transition to Idle (RFC 4271 §8.2.2 `OpenConfirm` + `HoldTimer_Expires` → send NOTIFICATION with Hold Timer Expired, release resources, Idle) rather than dropping the connection silently.

Worth handling the `MaxActiveSessions` cap from #265 in the same pass — the two together are what actually bounds the FD budget.

## Regression test

`BGPLite.Tests/BgpSessionHoldTimerTests.cs` already injects a `FakeTimeProvider`; add:

```csharp
[Fact]
public async Task OpenConfirm_PeerNeverSendsKeepalive_SessionTearsDownOnHoldTimer()
{
    // send OPEN only, advance the fake clock past the negotiated hold time
    timeProvider.Advance(TimeSpan.FromSeconds(negotiatedHoldTime + 1));
    await runTask.WaitAsync(TimeSpan.FromSeconds(5));   // must not hang
    Assert.Equal(BgpFsmState.Idle, session.State);
    var notif = Assert.Single(sent.OfType<BgpNotificationMessage>());
    Assert.Equal(BgpConstants.Error.HoldTimerExpired, notif.ErrorCode);
}
```

Refs: #115 (connect-to-OPEN timeout — bounds only the first read), #265 (max-active unenforced), #10 (FSM epic).

ruhex:
Closes #286

## Problem

`BgpConfig.OpenTimeoutSeconds` (#115) bounds only the **first** read — the one that receives the peer's OPEN. The KEEPALIVE read that follows it was unbounded:

```csharp
TransitionTo(BgpFsmState.OpenConfirm);
var response = await ReceiveMessageAsync(linkedCts.Token);   // no timer of any kind
```

`linkedCts` carries only the host shutdown token and the session's own `_cts`; neither fires on peer inactivity. `_negotiatedHoldTime` is assigned in `ValidateOpen`, but `HoldTimerLoopAsync` does not start until `RunEstablishedAsync` — i.e. after this read has already returned.

So a peer that sends a well-formed OPEN and then goes silent pins a `BgpSession`, a socket FD, a `NetworkStream` and a `Task` in `OpenConfirm` **forever**. Measured on `main` with `OpenTimeoutSeconds = 2`, `HoldTime = 3`:

```
after 20.0s: session state=OpenConfirm, RunAsync completed=False
-> UNBOUNDED: session + socket FD + task pinned in OpenConfirm forever
```

Both timers that should have fired are more than six windows past.

This walks straight past the #115 Slowloris defence — the attacker is not slow, it completes a legitimate OPEN exchange and stops. `MaxAcceptsPerIpPerMinute` (default 60) caps the accept *rate*, not the number of pinned sessions, so one source IP accumulates 60 per minute with no ceiling.

RFC 4271 §8.2.2 runs the Hold Timer in `OpenSent`/`OpenConfirm` as well; BGPLite ran no timer in either.

## Fix

Bound the KEEPALIVE read with the negotiated hold time, via a `CancellationTokenSource(TimeSpan, TimeProvider)` linked to `linkedCts` — the same pattern the OPEN timeout directly above already uses.

On expiry: `NOTIFICATION(Hold Timer Expired, 0)` and Idle, per RFC 4271 §8.2.2 (`OpenConfirm` + `HoldTimer_Expires`). Unlike the OPEN timeout — which deliberately drops the peer silently because a Slowloris socket would not read it — here we **do** notify: the peer completed the OPEN exchange, so it is reading the socket and the diagnostic reaches its operator.

The `when (!linkedCts.IsCancellationRequested)` filter distinguishes "the hold timer fired" from "we are shutting down", mirroring the OPEN-timeout handler.

### Hold time 0 — a deliberate deviation, flagged

RFC 4271 §4.2 disables the Hold Timer at 0. That is a rule for an **established** session; a handshake that never completes is not that, and leaving it unbounded would reopen exactly the hole this PR closes for any peer that proposes 0.

`OpenConfirm` therefore falls back to §8.2.2's initial Hold Time (suggested 4 minutes) when the negotiated value is 0. The Established phase still honors 0 as "no timer" — unchanged. A peer needing more than four minutes to send the KEEPALIVE that answers our own is not a case worth preserving.

No new config knob: the negotiated hold time is the RFC-correct source, and reusing `OpenTimeoutSeconds` (semantically "connect-to-OPEN") would be misleading.

## Verification

`dotnet build BGPLite.sln` + `dotnet test BGPLite.sln`: **602 passed, 0 failed** (600 before, 2 added). `dotnet format --severity warn` clean.

Both new tests were confirmed to catch the regression — with the bound removed they fail, with it restored they pass:

```
OpenConfirm bound temporarily removed
  Failed BgpSessionHoldTimerTests.OpenConfirm_HoldTimeZero_StillBoundedByFallback [5 s]
  Failed BgpSessionHoldTimerTests.OpenConfirm_PeerNeverSendsKeepalive_TearsDownOnHoldTimer [5 s]
--- restored ---
  Passed! - Failed: 0, Passed: 8
```

## Tests added

Both in `BgpSessionHoldTimerTests`, using the existing `FakeBgpConnection` + `FakeTimeProvider` seam (#96) — the timeout CTS is built on `_timeProvider`, so the clock advances instantly instead of waiting wall-clock minutes (248 ms for the whole file).

- `OpenConfirm_PeerNeverSendsKeepalive_TearsDownOnHoldTimer` — peer sends OPEN and nothing else; asserts `RunAsync` completes, state is `Idle`, and exactly one NOTIFICATION with `HoldTimerExpired`/`Unspecific` went out. `OpenTimeoutSeconds` is set to 30 — deliberately longer than `HoldTime` — so the test cannot pass by accident on the #115 timeout.
- `OpenConfirm_HoldTimeZero_StillBoundedByFallback` — asserts both halves of the fallback: still waiting at 3 minutes (behaviour unchanged inside the window), bounded past 4.

## Note

The companion gap — no cap on the number of concurrently active sessions — is a line item in #265 and is not addressed here (one issue : one PR). The two together are what actually bounds the FD budget: this PR stops a session being pinned forever, #265 would bound how many can exist at once.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented BGP sessions from remaining indefinitely in the OpenConfirm phase when a peer does not send a KEEPALIVE.
  - Added a four-minute fallback timeout when the negotiated hold time is zero.
  - Sessions now close cleanly and report a hold-timer expiration when the timeout is reached.

- **Tests**
  - Added coverage for missing KEEPALIVE messages and zero hold-time fallback behavior.

### DIFF
--- a/BGPLite.Server/BgpSession.cs
+++ b/BGPLite.Server/BgpSession.cs
@@ -73,6 +73,12 @@ public sealed class BgpSession : IDisposable
     // DoS where a peer spams type-5 and forces a full re-advertise. Initial 0 = never refreshed.
     // Read/written via Interlocked so RefreshRoutesAsync and ReadLoopAsync can't race.
     private long _lastRouteRefreshTicks;
+    // OpenConfirm bound when the negotiated hold time is 0 (#286). RFC 4271 §4.2 disables the Hold
+    // Timer at 0, but §8.2.2 also gives OpenSent a "large value" initial Hold Time with a suggested
+    // 4 minutes — a handshake that never completes is not the same thing as an established session
+    // that deliberately runs without timers, and leaving it unbounded is the resource hole this
+    // constant closes. The Established phase still honors hold time 0 as "no timer".
+    private static readonly TimeSpan OpenConfirmFallbackHoldTime = TimeSpan.FromMinutes(4);
     // Minimum gap between peer-triggered route refreshes. 1s is a reasonable default:
     // long enough to make flood-DoS impractical, short enough that a legitimate peer retry
     // after a lost UPDATE still gets a fresh advertisement promptly.
@@ -307,8 +313,44 @@ public sealed class BgpSession : IDisposable
 
             TransitionTo(BgpFsmState.OpenConfirm);
 
-            // Receive KEEPALIVE
-            var response = await ReceiveMessageAsync(linkedCts.Token);
+            // Receive KEEPALIVE, bounded by the OpenConfirm hold timer (#286). RFC 4271 §8.2.2 runs
+            // the Hold Timer in OpenSent/OpenConfirm as well, with the negotiated value once the
+            // OPEN exchange has happened. Without it this read was unbounded: #115's
+            // OpenTimeoutSeconds only covers the read that RECEIVES the OPEN, and the keepalive/hold
+            // loop does not start until RunEstablishedAsync — so a peer that sent a well-formed OPEN
+            // and then went silent pinned a session, a socket FD and a task indefinitely, walking
+            // straight past the Slowloris defence (it is not slow; it completes the OPEN and stops).
+            var confirmHoldTime = _negotiatedHoldTime > 0
+                ? TimeSpan.FromSeconds(_negotiatedHoldTime)
+                : OpenConfirmFallbackHoldTime;
+
+            BgpMessage response;
+            using (var confirmTimeoutCts = new CancellationTokenSource(confirmHoldTime, _timeProvider))
+            using (var confirmCts = CancellationTokenSource.CreateLinkedTokenSource(linkedCts.Token, confirmTimeoutCts.Token))
+            {
+                try
+                {
+                    response = await ReceiveMessageAsync(confirmCts.Token);
+                }
+                catch (OperationCanceledException) when (!linkedCts.IsCancellationRequested)
+                {
+                    // Only the OpenConfirm hold timer fired (the external/session token is still
+                    // alive). RFC 4271 §8.2.2, OpenConfirm + HoldTimer_Expires: send a NOTIFICATION
+                    // with Hold Timer Expired, release resources, go to Idle. Unlike the OPEN
+                    // timeout above we DO notify — the peer completed the OPEN exchange, so it is
+                    // reading the socket and the diagnostic reaches its operator.
+                    _logger.LogWarning(
+                        "Hold timer expired for {Peer} in OpenConfirm (no KEEPALIVE within {Hold}s) — closing (#286)",
+                        _peer, confirmHoldTime.TotalSeconds);
+                    if (Interlocked.CompareExchange(ref _teardownReason, (int)TeardownReason.HoldTimerExpired, (int)TeardownReason.None) == (int)TeardownReason.None)
+                    {
+                        try { await SendNotificationAsync(BgpConstants.Error.HoldTimerExpired, BgpConstants.SubError.Unspecific); }
+                        catch { /* best-effort — partial write counts, see RFC 4271 §8.1 */ }
+                    }
+                    return;
+                }
+            }
+
             _logger.LogInformation("Received {Type} from {Peer} in OpenConfirm", response.Type, _peer);
 
             switch (response)
--- a/BGPLite.Tests/BgpSessionHoldTimerTests.cs
+++ b/BGPLite.Tests/BgpSessionHoldTimerTests.cs
@@ -233,6 +233,133 @@ public class BgpSessionHoldTimerTests
     }
 
     /// <summary>
+    /// #286: a peer that sends a well-formed OPEN and then goes silent must NOT pin the session.
+    /// OpenTimeoutSeconds (#115) bounds only the read that receives the OPEN; the KEEPALIVE read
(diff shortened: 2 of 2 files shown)


ruhex:
@coderabbitai review

Issue #304, opened 2026-08-27
server: no max-prefix limit — one peer can grow the shared RouteTable without bound (RFC 4271 §6.7 / RFC 4486)


ruhex:
Split out of #292 (item 7). Verified against `main` @ 9708cdc.

## Problem

`BgpSession.HandleUpdateAsync` installs every accepted NLRI into the shared `RouteTable` with no ceiling:

```csharp
if (_routeFilter.AcceptIncoming(route, filterPeerConfig))
{
    _routeTable.AddOrUpdate(route);
    _installedByPeer.Add((nlri.Address, nlri.Length));
}
```

Measured on a live session: one peer announcing 4 000 prefixes across 20 UPDATEs grew the table to 4 000 entries with the session untouched. Nothing bounds it — a peer can announce until the process runs out of memory.

`BgpConstants.SubError.CeaseMaxPrefixes` (RFC 4486 subcode 1) has existed since #235 and is never used.

## Why it matters more after #289

#289 added `_installedByPeer`, a per-session `HashSet` that grows alongside the table. That is the right structure, but it means an unbounded announcement now costs memory in two places rather than one. The cap belongs with it.

## Recommended fix

- `BgpConfig.MaxPrefixesPerPeer` (0 = unlimited, matching the existing `MaxAcceptsPerIpPerMinute` / `OpenTimeoutSeconds` convention), validated at config load.
- When a peer's installed count would exceed it: `NOTIFICATION(Cease, CeaseMaxPrefixes)` and tear the session down, per RFC 4271 §6.7 and RFC 4486 §2.
- Consider a warning threshold (e.g. 75%) logged once per session — standard on real implementations and cheap here.

Counting against `_installedByPeer.Count` is the natural fit: it is already per-peer and already tracks exactly what this session put in the table.

## Test

```csharp
[Fact]
public async Task ExceedingMaxPrefixes_SendsCeaseAndTearsDown()
{
    // config MaxPrefixesPerPeer = 2, announce 3
    Assert.False(session.IsEstablished);
    var notif = Assert.Single(sent.OfType<BgpNotificationMessage>());
    Assert.Equal(BgpConstants.Error.Cease, notif.ErrorCode);
    Assert.Equal(BgpConstants.SubError.CeaseMaxPrefixes, notif.SubErrorCode);
}
```
plus a control asserting that a peer at exactly the limit stays up, and that `MaxPrefixesPerPeer = 0` is unlimited.

## Severity

[...]

Issue #289, opened 2026-08-26
routing: no Adj-RIB-In — any peer can withdraw routes it never announced from the shared RouteTable (RFC 4271 §3.2/§9)


ruhex:
## Summary

`RouteTable` is a single process-wide `ConcurrentDictionary<(uint Prefix, byte Length), Route>` shared by every `BgpSession` and by `RouteSeedingService`. `BgpSession.HandleUpdateAsync` applies a peer's withdrawals straight to it, keyed by prefix only:

```csharp
foreach (var w in update.WithdrawnRoutes)
{
    _routeTable.Remove(w.Address, w.Length);
    _logger.LogDebug("Route withdrawn: {Prefix}", w);
}
```

There is no Adj-RIB-In, no per-peer ownership, and no check that the withdrawing peer ever announced the prefix. RFC 4271 §3.2 models Adj-RIBs-In as *per peer*; §9 defines withdrawal as removing the route **received from that peer**.

## Impact

Any peer that completes a handshake — and BGPLite auto-registers unknown peers with RU defaults by default — can delete arbitrary entries from the shared table, including:

- prefixes seeded at startup by `RouteSeedingService` from the configured `PrefixSources`,
- routes announced by *other* peers.

Today the practical blast radius is bounded because in the production DI graph (`Program.cs` registers `PeerStore`, `IPrefixService` and `AppConfig`) `RouteAssembler.BuildOutboundRoutesAsync` always takes the peer-store branch and never reaches its shared-table fallback, so a deleted route is not immediately re-advertised. What *is* affected today:

- `GET /api/routes` (`ManagementApi.cs:1051` → `_routeTable.GetAll()`) — the operator-facing route view,
- `_metrics.SetRouteCount(_routeTable.Count)` — the exported route count,
- `RouteAssembler`'s shared-table fallback, which **is** the advertise path whenever `_peerStore`, `_prefixService` or `_appConfig` is null (all three are `IPeerStore?`/`IPrefixService?`/`AppConfig?` optional constructor parameters — see #263 for how easily that happens).

In that last configuration this is a straightforward cross-peer route-injection/withdrawal primitive.

## Reproduction

Seed two routes, connect one peer, have it send an UPDATE with **no NLRI and no attributes** — purely withdrawals for prefixes it never announced:

```
routeTable before=2 after=0 established=True
```

[...]

Issue #263, opened 2026-08-26
architecture: nullable-optional dependencies across Server/Api/Providers — a missing DI registration silently disables features


ruhex:
## What is wrong

Nearly every core dependency is declared nullable-optional:
- `BgpServer.cs:20-24,50-54` — `IPeerStore?`, `IPrefixService?`, `IPrefixAggregator?`, `ICommunityResolver?`;
- `BgpSession.cs:161-175` — 14 constructor parameters, 7 optional;
- `RouteAssembler.cs:22-24` — `IPrefixService?`, `IPeerStore?`, `AppConfig?`;
- `ManagementApi.cs:57-59` — `IPrefixService?`, `IPrefixSourceService?`, `ISessionManager?`;
- `PrefixService.cs:14-15,41-42` — `RipeStatProvider?`, `HttpPrefixProvider?`.

Fallback behavior is silent: `_httpProvider is null → return []` (`PrefixService.cs:64`), `NullCommunityResolver.Instance` (`BgpServer.cs:66`), `_sessionManager is not null` guards (`ManagementApi.cs:664`). `Program.cs` currently registers everything, but the type contract does not enforce it — one removed registration (or a hand-rolled composition like `BgpServer.cs:194`, `new BgpSession(...)` with 11 args) turns into "why are no prefixes announced" instead of a startup failure.

## Impact

Composition errors are undiagnosable at runtime; the send path is littered with maybe-null handling for configuration that in production is always present.

## Fix

1. Make production dependencies non-nullable and move "absence" into explicit DI null-object registrations in `Program.cs` (e.g. `AddSingleton<ICommunityResolver>(NullCommunityResolver.Instance)` when unconfigured) — tests keep using optional parameters in the meantime.
2. Introduce a `BgpSessionFactory` (and let RouteAssembler be injected) so `new BgpSession`/`new RouteAssembler` leave `BgpServer`/`BgpSession`.

Each step independently compiles and is covered by existing tests.

## Acceptance

- Removing any single DI registration fails loudly at startup (or resolves an explicit null object that logs its activation once).
- `BgpServer`/`BgpSession` no longer `new` up their collaborators.

Refs: audit 2026-08-26; #88/#98/#105 (architecture-review cluster).

ruhex:
## Severity is understated for one specific instance — it is not "features silently disabled", it is "wrong routes served"

[...]

PR #308, opened 2026-08-27
fix(routing): the shared-table fallback must not advertise peer-injected routes

### LINKED ISSUE #307: routing,security: the RouteAssembler fallback advertises peer-injected routes — one missing DI registration crosses the tenant boundary
Split from #263, which covers the general nullable-dependency problem. This is its one instance where a missing registration serves the **wrong routes** rather than no routes, and it crosses a tenant boundary.

## What is wrong

`RouteAssembler.BuildOutboundRoutesAsync` takes the per-peer decision tree only when all three optional dependencies are present, and otherwise advertises the shared `RouteTable` verbatim:

```csharp
if (_peerStore is not null && _prefixService is not null && _appConfig is not null)
{
    // subscriptions / custom prefixes / custom ASNs / user URL sources
    ...
}

// Final fallback: send from shared route table.
foreach (var r in _routeTable.Enumerate())
    if (_routeFilter.AcceptOutgoing(r, filterPeerConfig, sharedAllowSet))
        filtered.Add(r);
```

All three arrive as optional constructor parameters defaulting to `null` — `RouteAssembler.cs:35-39`, reached through `BgpSession.cs:246-248` and `BgpServer.cs:52-53`.

The shared `RouteTable` is not a per-peer route set. It holds the startup seed from `RouteSeedingService` **and every NLRI any peer has announced inbound** — `BgpSession.HandleUpdateAsync` writes accepted routes straight into it.

## Impact

BGPLite serves each user the prefix set that user selected in the management UI. With one registration missing, every session instead receives one global list that includes routes other users' sessions injected. No exception, no distinguishing log line, sessions establish normally — the failure is invisible from outside.

That is a tenant-isolation failure, not a disabled feature: a user can put a prefix into another user's advertisement by announcing it inbound.

The fallback is unreachable in the current `Program.cs`, which registers all three. Nothing in the type system keeps it that way, which is exactly #263's point; the difference here is the blast radius when it does happen.

## Fix

Two parts, both small:

1. **The fallback must never re-advertise peer-injected routes.** #289 added an owner tag to every `RouteTable` entry: the startup seed is stored with a null owner, peer-installed routes with the owning session. So the fallback can enumerate only unowned entries and the cross-tenant path closes outright, independent of DI:

   ```csharp
   public IEnumerable<Route> EnumerateUnowned()   // startup seed only
   ```

2. **Make it audible.** Log once per session at Warning when the fallback branch is taken, naming which dependency is missing. Today the only symptom is "the peer got the wrong prefixes".

Deliberately not a throw: several tests legitimately construct `BgpSession` without a peer store and rely on this path. Making it fail loudly at composition time is #263's fix (1) and stays there.

## Acceptance

- A peer's inbound announcement can never appear in another peer's advertisement, including on the fallback path.
- Taking the fallback logs a warning that names the missing dependency.
- Existing tests that construct `BgpSession` without a peer store keep working against seeded routes.

## Test

```csharp
[Fact]
public async Task Fallback_DoesNotAdvertiseRoutesInjectedByAnotherPeer()
{
    // seed one route; have peer A announce another; assert peer B's advertisement
    // contains the seeded prefix and NOT A's.
}
```

Refs #263 (parent), #289 (the owner tag this builds on), #304 (nothing bounds what a peer may add to that table).

ruhex:
Closes #307

## Problem

`RouteAssembler.BuildOutboundRoutesAsync` takes the per-peer decision tree only when all three of `IPeerStore`, `IPrefixService` and `AppConfig` are present — each an optional constructor parameter defaulting to `null` (`RouteAssembler.cs:35-39`, reached via `BgpSession.cs:246-248`, `BgpServer.cs:52-53`) — and otherwise advertises the shared `RouteTable` verbatim.

That table is not a per-peer route set. Alongside the startup seed from `RouteSeedingService`, it holds **every NLRI any peer announced inbound**, because `BgpSession.HandleUpdateAsync` installs accepted routes straight into it.

So one missing DI registration silently changes what every session receives, from *"the prefixes this user selected"* to *"one global list including what other users' sessions injected"*. No exception, no distinguishing log line, sessions establish normally.

BGPLite is deployed as a service where each user picks their own prefix set in the management UI and the session exists to hand that user their selection. Under that model this is a **tenant-isolation failure**, not a disabled feature: a peer can place a prefix into another peer's advertisement simply by announcing it inbound.

The branch is unreachable in the current `Program.cs`, which registers all three. Nothing in the type system keeps it that way — that is #263's subject; what this PR addresses is the blast radius when it does happen.

## Fix

**1. `RouteTable.EnumerateUnowned()`** yields only entries installed with no owner — in practice the startup seed. The owner tag added in #289 already separates those from peer-installed routes, so the fallback stops being a cross-tenant path *outright*, rather than depending on the composition being correct:

```csharp
foreach (var r in _routeTable.EnumerateUnowned())   // was: Enumerate()
```

This is the advertise-side counterpart of `RemoveOwnedBy`. The tag now carries one coherent meaning in both directions: a peer can neither delete what it does not own, nor have what it injected handed to somebody else.

**2. Taking the fallback logs a Warning** naming which dependency is missing. Until now the only symptom was "the peer got the wrong prefixes", which is indistinguishable from a configuration mistake on the user's side.

## Deliberately not a throw

Several existing tests construct `BgpSession` without a peer store and legitimately exercise this path (`BgpSessionShutdownTests`, `MalformedUpdateResilienceTests`, `SendTimeoutTests`, `BgpSessionHoldTimerTests`, and the ownership tests here). Turning the fallback into a failure would need those to be given a null-object to inject first — that is #263's fix (1), and it belongs there. This PR closes the isolation hole independently of whether that lands, which seemed the right order given the two have very different sizes.

## Verification

`dotnet build BGPLite.sln` + `dotnet test BGPLite.sln`: **684 passed, 0 failed** (683 before, 1 added). `dotnet format --severity warn` clean. No existing test changed behaviour — the two that seed the table already use the owner-less `AddOrUpdate(route)` overload, which is exactly what the fallback still serves.

Confirmed to catch the regression — reverting `EnumerateUnowned()` to `Enumerate()`:

```
Failed SharedTableFallback_DoesNotAdvertiseRoutesInjectedByAnotherPeer
  Assert.DoesNotContain() Failure: Item found in collection
```

## Test added

`SharedTableFallback_DoesNotAdvertiseRoutesInjectedByAnotherPeer` — seeds `192.0.2.0/24`, has one peer announce `10.0.0.0/8` inbound (asserting it *is* in the shared table, so the test cannot pass by the injection having failed), then connects a second peer and reads the NLRI it actually put on the wire. The seed must be there; the other peer's injection must not.

`ScriptedConnection` now captures outbound frames so that last assertion is against real wire bytes rather than internal state.

## Note

Nothing bounds what a peer may *add* to that table — see #304. This PR stops those additions leaking into other peers' advertisements; it does not stop them accumulating.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected fallback route announcements so peers receive only configured startup routes.
  * Prevented routes learned from other peer sessions from being advertised during an initial dump.
  * Added a warning when peer-specific configuration is unavailable and configured prefixes cannot be served.

* **Tests**
  * Added coverage verifying route ownership and outbound route announcements.

### DIFF
--- a/BGPLite.Routing/RouteTable.cs
+++ b/BGPLite.Routing/RouteTable.cs
@@ -95,6 +95,26 @@ public sealed class RouteTable
             yield return entry.Route;
     }
 
+    /// <summary>
+    /// Enumerates only the routes installed with no owner — in practice the startup seed written by
+    /// <c>RouteSeedingService</c>. Excludes everything a peer announced inbound, which
+    /// <c>BgpSession.HandleUpdateAsync</c> installs owned by its session.
+    /// <para>
+    /// This is the advertise-side counterpart to <see cref="RemoveOwnedBy"/>: the owner tag exists so
+    /// a peer can neither remove what it does not own nor have what it injected handed to somebody
+    /// else. Used by <c>RouteAssembler</c>'s shared-table fallback, which would otherwise re-advertise
+    /// one peer's inbound announcements to every other peer (#307).
+    /// </para>
+    /// </summary>
+    public IEnumerable<Route> EnumerateUnowned()
+    {
+        foreach (var entry in _routes.Values)
+        {
+            if (entry.Owner is null)
+                yield return entry.Route;
+        }
+    }
+
     public void Clear() =>
         _routes.Clear();
 }
--- a/BGPLite.Server/RouteAssembler.cs
+++ b/BGPLite.Server/RouteAssembler.cs
@@ -268,10 +268,23 @@ internal sealed class RouteAssembler
             }
         }
 
-        // Final fallback: send from shared route table.
+        // Fallback: no per-peer configuration is reachable, so serve the shared table's SEEDED
+        // routes. This branch means a dependency is missing from the composition — in a correct
+        // production wiring it is unreachable (#263), and reaching it silently changes what every
+        // peer receives rather than merely disabling a feature, so it is logged (#307).
+        _logger.LogWarning(
+            "Route assembly for {Peer} fell back to the shared table — per-peer configuration is unavailable " +
+            "(peerStore={HasPeerStore}, prefixService={HasPrefixService}, appConfig={HasAppConfig}). " +
+            "This peer will NOT receive its configured prefixes.",
+            _peer, _peerStore is not null, _prefixService is not null, _appConfig is not null);
+
+        // EnumerateUnowned, not Enumerate: everything a peer announced inbound is installed in this
+        // same table owned by its session (#289). Advertising those here would hand one peer's
+        // injected routes to every other peer — a tenant-isolation failure, not just a wrong list.
+        // The startup seed is written with no owner and is what this fallback is meant to serve.
         var sharedAllowSet = _routeFilter.ResolveOutgoingAllowSet(filterPeerConfig);
         var filtered = new List<Route>();
-        foreach (var r in _routeTable.Enumerate())
+        foreach (var r in _routeTable.EnumerateUnowned())
         {
             if (_routeFilter.AcceptOutgoing(r, filterPeerConfig, sharedAllowSet))
                 filtered.Add(r);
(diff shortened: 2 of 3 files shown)


ruhex:
@coderabbitai review

Issue #228, opened 2026-08-03
api: N+1 DbContexts in BuildPeerDetail / HandleGetPeer — 5 SQLite connections per peer on /api/me


ruhex:
## What is wrong

`ManagementApi.BuildPeerDetail` (the per-peer helper behind `/api/me`) opens **five separate `DbContext` instances** to build one peer-detail response:

```csharp
// BGPLite.Api/ManagementApi.cs:534-571
private object? BuildPeerDetail(string peerId)
{
    var peer = _store.GetDbPeerById(peerId);                 // DbContext #1
    if (peer is null) return null;
    var subscriptions = _store.GetSubscriptions(peer.Id);    // DbContext #2
    var customPrefixes = _store.GetCustomPrefixes(peer.Id);  // DbContext #3
    var customAsns = _store.GetCustomAsns(peer.Id);          // DbContext #4
    var communities = peer.Asn.HasValue
        ? _store.GetCommunities(peer.Ip, peer.Asn.Value)     // DbContext #5
        : new HashSet<uint>();
    ...
}
```

`HandleGetPeer` (`ManagementApi.cs:631-661`) has the same shape (six contexts). And `/api/me` calls `BuildPeerDetail` in a `Select` over **every** peer on the resolved client IP (`ManagementApi.cs:529`).

**Impact:**
- For N peers sharing one source IP (NAT/VPN — the exact scenario `/api/me` exists for, see #23), one `/api/me` call opens **5N `DbContext` instances**. Each `DbContext` = a new SQLite connection + the 3 `PRAGMA` round-trips from `SqlitePragmasInterceptor`. That is `5N × (connect + 3 PRAGMA + 4 queries)` per request.
- This is especially wasteful because the codebase **already solved** this exact problem in `PeerStore.LoadPeerRoutingView` (`PeerStore.cs:126-154`), which loads the peer + four child collections in a **single** `AsNoTracking` query — explicitly documented as "six statements on five connections → two statements on one" (issue #84). But `BuildPeerDetail` / `HandleGetPeer` do not use it.
- All `PeerStore` methods are synchronous, so each connection also blocks a thread-pool thread for the duration of the SQLite I/O (minor on SQLite, but inconsistent with the async-everywhere pattern).

## Fix

[...]

Issue #138, opened 2026-07-05
[perf] Suppress EF Core MultipleCollectionIncludeWarning (SingleQuery OK for SQLite + small data)


ruhex:
EF Core emits `warning 20504` (MultipleCollectionIncludeWarning) on the `LoadPeerRoutingView` query (#84) because it loads 3 collection navigations (Subscriptions, CustomPrefixes, CustomAsns) in a single query.

## Assessment (after CodeRabbit review)
**AsSplitQuery was tried first (PR #139 v1) but reverted** — it makes the read non-atomic (3 separate SELECTs can see mixed old/new state during concurrent writes, per CodeRabbit 🟡). For SQLite (local, embedded, ~0ms latency) with small per-peer data (tens of rows), the Cartesian product from SingleQuery is negligible.

## Fix (industry best practice for this scenario)
Suppress the warning globally via `ConfigureWarnings(w => w.Ignore(RelationalEventId.MultipleCollectionIncludeWarning))` on both DbContext registrations. Keeps SingleQuery (atomic read) + eliminates the noisy warning. Documented with rationale (small data, local SQLite, atomicity matters).

**When AsSplitQuery WOULD be the right choice:** large collections (1000+ rows) or remote DB (SQL Server/PostgreSQL with network latency). Not the case here.

Issue #95, opened 2026-07-03
[arch/major] resilience: Every PeerStore mutation (CreatePeer, UpsertPeer, Updat


ruhex:
From the architecture + .NET 10/C# best-practices review (2026-07-03, 12 aspect-agents, 83 findings).

## Problem (MAJOR, high)
**`BGPLite.Api/PeerStore.cs:13-205`** (resilience-errors)

Every PeerStore mutation (CreatePeer, UpsertPeer, UpdateSessionStatus, SetCommunities, SetSubscriptions, etc.) calls SaveChanges/ExecuteUpdate with no retry, no timeout, and no exception handling. SQLite is file-locked under concurrent writers (the management API, per-session UpsertPeer on OPEN, UpdateSessionStatus on session up/down all hit the same file). A transient SQLITE_BUSY / 'database is locked' throws straight out. UpsertPeer is called from the session's _onPeerIdentified callback (Program.cs:119) — a locked DB throws during OPEN handling and will surface in the session catch-all as a Cease. SetCommunities/SetSubscriptions throw out of the management-API handler into the generic 500 path, returning 'database is locked' to the HTTP client with no retry. WAL is not enabled (Program.cs:37-41 sets only Data Source), so writer contention is high.

## Fix
Enable WAL + busy_timeout at the connection level (PRAGMA journal_mode=WAL; PRAGMA busy_timeout=5000) in the DbContext options or OnConfiguring, and add a retry policy (Polly or a small SQLite-busy retry loop) around writes in PeerStore. Treat a truly unwritable DB (disk full, read-only FS) as fail-loud at startup with a clear log, not a per-request 500.

PR #309, opened 2026-08-27
perf(api): split the peer reads — a Cartesian product was costing up to 814 ms per call

### LINKED ISSUE #260: api: LoadPeerRoutingView cartesian explosion — 4 collection Includes in one query on the BGP send path
## What is wrong

`LoadPeerRoutingView` (`PeerStore.cs:174-179`) eagerly Includes Subscriptions + CustomPrefixes + CustomAsns + filtered CustomSources in a **single** query (SingleQuery; the warning is suppressed in `Program.cs:51-56`). The justification (#138, "tens of rows") does not hold for API-loaded lists: request bodies up to 64 MB mean hundreds of thousands of custom prefixes are possible.

## Impact

The JOIN produces |Subs|×|Prefixes|×|Asns|×|Sources| rows streamed through the SQLite reader on every session establish / refresh — e.g. 500 prefixes × 50 ASN × 10 sources × 5 subs = 1.25M rows. This runs on the hot send path (`RouteAssembler.cs:75`) and is fully synchronous (see the async-PeerStore issue), so it also blocks a thread-pool thread for the duration.

## Fix

Project into a DTO the way `GetPeerDetail` already does (`PeerStore.cs:220-239` — EF generates split queries per collection automatically), or use `AsSplitQuery()` inside one transaction when snapshot atomicity is required.

## Acceptance

- Row count scales as sum-of-collections, not product; a peer with large custom lists no longer multiplies query cost.
- LoadPeerRoutingView result is byte-identical to today (regression test with all four collections populated).

Refs: audit 2026-08-26; #138, #228.

ruhex:
Closes #260

## Measured, not argued

Both peer reads emitted **one** statement LEFT JOINing every child collection, so the driver materialized their Cartesian product. Real SQLite file, 200 iterations per shape:

| peer shape | cartesian rows | before | after | |
|---|---|---|---|---|
| 1 sub, 5 prefixes, 1 asn, 1 source | 5 | 0.24 ms | 0.27 ms | — |
| 3 subs, **200 prefixes**, 5 asns, 2 sources | 6,000 | **31 ms** | **0.33 ms** | 96× |
| 5 subs, **1000 prefixes**, 10 asns, 3 sources | 150,000 | **814 ms** | **1.4 ms** | 596× |

`LoadPeerRoutingView` is on the BGP send path — every session establish, every `RefreshRoutesAsync`, and `RefreshAllEstablishedAsync` fires it for all peers concurrently. A user with 1,000 custom prefixes waited most of a second on the database before route assembly even began.

`GetPeerDetail` is the read the management UI hits to render a peer: **120 ms → 0.17 ms** at 200 prefixes.

## Two prior conclusions this corrects

Both verified by dumping the SQL EF actually emits, not by reading the code.

**#138's premise is false.** It suppressed `MultipleCollectionIncludeWarning` because *"for SQLite (local, embedded) with small per-peer data (tens of rows), the Cartesian product is negligible and SingleQuery keeps the read atomic"*. The product is `subs × prefixes × asns × sources`, and this deployment lets a user pick prefixes in a UI — 200 is ordinary, and six figures needs nothing unusual. The suppression is removed: with both reads split it has nothing to fire on, and leaving it would only hide the next reintroduction.

**#228's fix did not do what its doc claims.** `GetPeerDetail` was documented as:

> EF auto-splits the collection subqueries inside a projection (one SELECT per collection + one for the peer row, all on the same connection), so there is no Cartesian-product row explosion that an Include-based load would produce.

EF emits a single joined statement for it — byte-for-byte the same shape as the `Include` version:

```sql
SELECT "p4"."Id", "p0"."AsnListName", ... FROM (
    SELECT "p"."Id" FROM "Peers" AS "p" WHERE "p"."Ip" = ... LIMIT 1
) AS "p4"
LEFT JOIN "PeerSubscription" AS "p0" ON "p4"."Id" = "p0"."PeerId"
LEFT JOIN "PeerCustomPrefix" AS "p1" ON "p4"."Id" = "p1"."PeerId"
...
```

**One** SELECT before, **six** after adding `AsSplitQuery`. So the GET path #228 set out to fix has carried the explosion since — with five collections, one more than `LoadPeerRoutingView`. The doc is corrected in place rather than deleted, since the false claim is exactly the thing a future reader would otherwise rely on.

This also means the obvious reading of #260 — "switch the send path to a projection like `GetPeerDetail`" — would have produced a 1.2× improvement and closed the issue as fixed. That is why the measurement came before the change.

## Consistency

Splitting gives up the single-statement read consistency #138 also named — the one thing it got right. Both reads therefore run inside an explicit transaction: the separate SELECTs would otherwise each see their own snapshot, and a UI edit landing between them would advertise a mixed configuration. SQLite runs in WAL here (#95), so a read transaction takes no write lock and blocks no writer.

In `LoadPeerRoutingView` the transaction covers the read only; the folded `ExecuteUpdate` status write stays outside it, so the write lock is held no longer than before.

## Verification

`dotnet build BGPLite.sln` + `dotnet test BGPLite.sln`: **687 passed, 0 failed** (684 before, 3 added). `dotnet format --severity warn` clean.

Confirmed to catch the regression — with `AsSplitQuery()` removed:

```
Failed LoadPeerRoutingView_EmitsOneSelectPerCollection
  expected a split query (one SELECT per collection), got 1 SELECT statement(s)
Failed GetPeerDetail_EmitsOneSelectPerCollection
  expected a split query (one SELECT per collection), got 1 SELECT statement(s)
```

## Tests added

`PeerStoreSplitQueryTests` asserts the **emitted SQL** rather than timings — deterministic, no dependence on the runner, and it covers what EF's own `MultipleCollectionIncludeWarning` cannot: that warning fires only for the `Include` shape, so it would never have caught `GetPeerDetail`'s projection.

- `LoadPeerRoutingView_EmitsOneSelectPerCollection`
- `GetPeerDetail_EmitsOneSelectPerCollection`
- `SplitReadsReturnTheSameData` — the split must not change what the reads return, including that only Active sources reach the advertisement (#147) while the detail view shows both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer detail and routing views to load related information consistently.
  * Prevented duplicate or incomplete peer data when multiple related collections are displayed.
  * Preserved subscriptions, prefixes, ASNs, and source visibility during peer reads.

* **Tests**
  * Added coverage to verify accurate results and reliable loading of related peer information.

### DIFF
--- a/BGPLite.Api/PeerStore.cs
+++ b/BGPLite.Api/PeerStore.cs
@@ -171,12 +171,33 @@ public sealed class PeerStore : IPeerStore
     public PeerRoutingView? LoadPeerRoutingView(string ip, uint asn)
     {
         using var db = _dbFactory.CreateDbContext();
-        var peer = db.Peers.AsNoTracking()
-            .Include(p => p.Subscriptions)
-            .Include(p => p.CustomPrefixes)
-            .Include(p => p.CustomAsns)
-            .Include(p => p.CustomSources.Where(c => c.Active))
-            .FirstOrDefault(p => p.Ip == ip && p.Asn == asn);
+
+        // AsSplitQuery: without it EF emits ONE statement that LEFT JOINs all four collections, so
+        // the driver materializes subs x prefixes x asns x sources rows. Measured on a real SQLite
+        // file, 200 iterations: a peer with 3 subscriptions / 200 custom prefixes / 5 ASNs /
+        // 2 sources produces 6,000 rows and takes 31 ms per call; 5/1000/10/3 produces 150,000 rows
+        // and takes 814 ms. Split, the same reads are 0.33 ms and 1.4 ms — 96x and 596x (#260).
+        //
+        // This runs on the BGP send path (every session establish, every RefreshRoutesAsync, and
+        // RefreshAllEstablishedAsync fires it for all peers at once), so it is the read a user waits
+        // on before their selected prefixes reach the wire.
+        //
+        // The read is wrapped in a transaction because splitting gives up the single-statement
+        // consistency #138 relied on: the four SELECTs would otherwise each see their own snapshot,
+        // and a UI edit landing between them would advertise a mixed configuration. SQLite runs in
+        // WAL here (#95), so a read transaction takes no write lock and does not block writers.
+        Peer? peer;
+        using (var read = db.Database.BeginTransaction())
+        {
+            peer = db.Peers.AsNoTracking()
+                .Include(p => p.Subscriptions)
+                .Include(p => p.CustomPrefixes)
+                .Include(p => p.CustomAsns)
+                .Include(p => p.CustomSources.Where(c => c.Active))
+                .AsSplitQuery()
+                .FirstOrDefault(p => p.Ip == ip && p.Asn == asn);
+            read.Commit();
+        }
         if (peer is null) return null;
 
         // Fold the status update (was UpdateSessionStatus(active:true) on its own DbContext) into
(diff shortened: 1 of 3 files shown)


ruhex:
@coderabbitai review

Issue #300, opened 2026-08-26
protocol: RFC 4271 §6.1 header validation gaps + AS 0 accepted (RFC 7607) — split from #292


ruhex:
Split out of #292 (items 2, 4, 5) — they form one coherent change across `BgpMessageReader` / `OpenNegotiator` / `AttributeHelper`, where the rest of that issue is perf, dependency and policy work. All three verified against `main` @ 1224144 with a standalone harness.

## 1. KEEPALIVE with a length other than 19 is accepted (RFC 4271 §6.1)

`ReadMessage` maps type 4 straight to `BgpKeepaliveMessage.Instance` with no length check, so trailing bytes are silently ignored:

```
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF 0017 04 DEADBEEF   -> parsed as Keepalive, payload ignored
```

RFC 4271 §6.1:

> ... if the Length field of the message header is less than 19 or greater than 4096, or if the Length field of an OPEN message is less than the minimum length of the OPEN message, or if the Length field of an UPDATE message is less than the minimum length of the UPDATE message, or if the Length field of a KEEPALIVE message is not equal to 19, or if the Length field of a NOTIFICATION message is less than the minimum length of the NOTIFICATION message, then the Error Subcode MUST be set to Bad Message Length.

Only the global 19..4096 bound is enforced today. The per-type minimums for OPEN (29) / UPDATE (23) / NOTIFICATION (21) happen to be covered incidentally by their body-length checks — but as *body* errors with Update/Open Message Error, not as header errors. KEEPALIVE has no check at all. `BgpConstants.MinOpenMessageSize` exists and is unused.

## 2. Header errors carry Unspecific and no Data field (RFC 4271 §6.1)

```csharp
// BgpMessageReader.ReadMessage
_ => throw new BgpParseException($"Unknown message type: {type}")        // -> 1/0, no data

// BgpSession.ReceiveMessageAsync
if (length is < MinMessageSize or > MaxMessageSize)
    throw new BgpParseException($"Invalid message length: {length}");     // -> 1/0, no data
```

RFC 4271 §6.1 mandates **Bad Message Length (2)** with the erroneous Length in the Data field, and **Bad Message Type (3)** with the erroneous type in the Data field. Both currently emit `MessageHeaderError/Unspecific` with no Data, so the peer's operator gets "unknown error" and no diagnostic.

Verified:

```
[...]

Issue #266, opened 2026-08-26
chore: api audit minors — format=txt export, CORS PATCH, community/subscription-name validation, log sanitization gaps, MaxRequestBodyBytes reload, OPTIONS pre-rate-limit, misleading 409


ruhex:
A grouping issue for LOW/MINOR management-API findings from the 2026-08-26 audit.

## 1. `format=txt` export is broken by double serialization
`ManagementApi.cs:899-900` sets `text/plain`, but `WriteResponse` (`:1260-1264`) unconditionally overrides to `application/json` and JSON-encodes the string (quotes + `\\n`). Fix: don't overwrite an explicitly set ContentType; write txt bytes directly.

## 2. CORS Allow-Methods misses PATCH
`ManagementApi.cs:1303` lists GET/POST/PUT/DELETE/OPTIONS while a PATCH route exists (`:397-398`) — browser UI cannot toggle source active. Fix: add PATCH.

## 3. Community not validated on AddSource
`ManagementApi.cs:830` accepts any string; a typo later silently falls back to an auto-gen community at send time (`ConfigCommunityResolver.cs:103-114`). Fix: validate with a TryParse wrapper → 400.

## 4. Subscription list names not checked against config
`ManagementApi.cs:743` stores unknown names (e.g. "ru-backup") — silent no-op forever. Fix: 400/422 for names not in `RipeStat.AsnLists` / sources.

## 5. Log sanitization gaps
`ManagementApi.cs:639-641` (CustomAsns join unsanitized — neighbors sanitized), `:661-662` (`Created peer {Ip}` with raw `data.Ip`). Fix: SanitizeForLog consistently (moot for Ip once input validation lands).

## 6. MaxRequestBodyBytes is not hot-reloaded
`ApplyConfig` (`ManagementApi.cs:98-122`) swaps 4 fields but not `_config`, despite the doc comment claiming it; `ReadBodyAsync` reads `_config.MaxRequestBodyBytes` (`:443`) forever from startup. Fix: swap `_config` via Interlocked + Volatile reads, or fix the comment and document the field as restart-required.

## 7. OPTIONS preflight served before rate limiting
`ManagementApi.cs:265-270` returns before the rate-limit check at `:273` — unlimited cheap preflight flood (bounded only by the 64 in-flight slots). Fix: rate-limit OPTIONS too or move the check earlier.

## 8. Misleading 409 on concurrently-deleted peer
[...]

Issue #259, opened 2026-08-26
api: CreatePeer/UpdatePeer not atomic (4 DbContexts / 3 transactions); duplicate CIDR → 500 after partial commit


ruhex:
## What is wrong

`HandleCreatePeer` chains `CreatePeer → SetSubscriptions → SetCustomPrefixes → SetCustomAsns` (`ManagementApi.cs:654-657`; UpdatePeer same at `:742-748`). Each `Set*` opens its own DbContext/transaction (`PeerStore.cs:313-323, 335-345, 357-367`).

Additionally the parsed custom-prefix list is not deduplicated (`ManagementApi.cs:643-652`): `["10.0.0.0/8","10.0.0.0/8"]` violates the `(PeerId, Prefix, PrefixLength)` PK inside `SetCustomPrefixes`.

## Impact

- Duplicate CIDR → `SqliteException` → **500** to the client, but the peer row and subscriptions are already committed → partially configured peer; client retry creates conflict/garbage.
- Between `SetSubscriptions` and `SetCustomPrefixes`, a concurrent BGP session reads an incomplete `LoadPeerRoutingView` → announces without the custom prefixes.

## Fix

- Dedup prefixes (`Distinct` by `(Prefix, Length)`) before any mutation.
- Single DbContext + single transaction for the whole create/update (e.g. `PeerStore.CreatePeerFull(...)`), or at minimum all mutations before any commit point.

## Acceptance

- A create/update request either fully applies or applies nothing (no partial peer).
- Duplicate CIDRs are accepted idempotently (dedup) or rejected 400 **before** anything is persisted.

Refs: audit 2026-08-26; #226/#227 (per-set atomicity that this violates at the composition level).

Issue #156, opened 2026-07-05
[security/api] Unbounded request body + collection sizes — OOM DoS


ruhex:
## Bug

Every POST/PUT/PATCH handler in `ManagementApi` reads the body with no size cap and deserializes without collection limits:

```csharp
// BGPLite.Api/ManagementApi.cs:427, 515, 591, 625 (and similar in all mutation handlers)
using var reader = new StreamReader(ctx.Request.InputStream, ctx.Request.ContentEncoding);
var body = await reader.ReadToEndAsync();     // ❌ no ContentLength64 check, no cap
var data = JsonSerializer.Deserialize<PeerInput>(body);
// then: data.CustomPrefixes / CustomAsns / Lists / Sources are uncapped collections
```

## Impact

- **Memory DoS**: a single client can stream gigabytes into the process before deserialization even runs. `HttpListener` has no default body cap.
- **Collection DoS**: even a small JSON body can declare millions of `CustomPrefixes` / `CustomAsns` / `Subscriptions` / `Sources` per peer, which then get persisted and re-loaded on every peer refresh.
- Compounds with the absence of authentication (#90): anyone reachable can drive the process to OOM.

## Fix

- Reject requests where `ctx.Request.ContentLength64` exceeds a configured max (default e.g. 1 MiB) with `413 Payload Too Large`.
- Wrap `StreamReader` with a `BoundedStream` that also caps chunked-transfer bodies (no Content-Length).
- Enforce per-collection max counts (e.g. 10 000 custom prefixes, 1 000 ASNs) before mutating; return `400` with a clear message.
- Add a `maxRequestBodyBytes` / per-collection limits to `ApiConfig`.

## Scope

M. API-layer; covered by `ManagementApiTests` (or a new `RequestBodyLimitsTests`) sending oversized bodies and oversized collections.

PR #311, opened 2026-08-27
fix(api): validate and canonicalize peer input before it is persisted

### LINKED ISSUE #255: api: POST/PUT /api/peers accepts unvalidated Ip/Asn — phantom peers, ASN 0/AS_TRANS, 500 on missing fields
## What is wrong

`HandleCreatePeer` passes `data.Ip`/`data.Asn` straight to the store (`ManagementApi.cs:654`); there is no validation in the handler or in `PeerStore.CreatePeer` (`PeerStore.cs:13`). AGENTS.md explicitly requires: "API endpoints must validate peer addresses, ASN ranges".

Concrete vectors:
- `"ip": "1.2.3.4 "` (trailing space) or a non-canonical IPv4-mapped form: accepted with 200; the BGP path later writes the canonical `remoteEndpoint.Address.ToString()` (`BgpServer.cs:172`), the `(Ip, Asn)` upsert does not match → **phantom second peer row; the configured peer never binds to its session**.
- `"ip": "banana"`, arbitrary-length strings (up to the 64 MB body cap), `asn: 0` (invalid per RFC 7607), `asn: 23456` (AS_TRANS) — all silently accepted.
- JSON without `"ip"` → `Ip = null` → SQLite NOT NULL violation → **500** "Internal server error" instead of 400 (`MapExceptionToResponse` does not distinguish).

## Fix

In `HandleCreatePeer`/`HandleUpdatePeer`: `IPAddress.TryParse` + `AddressFamily == InterNetwork` + canonicalize (reject IPv4-mapped), reject AS 0 (and optionally 23456 / 65535 / 4200000000+), require `Ip` (null-check before the store), cap string lengths.

## Acceptance

- `banana` / `asn:0` / missing `ip` → 400 with a specific message; nothing persisted.
- A peer created with a valid IP binds to its BGP session (no phantom rows).

Refs: audit 2026-08-26.

ruhex:
Closes #255

## Problem

`HandleCreatePeer` passed `data.Ip` and `data.Asn` straight to the store, and both handlers passed `data.CustomAsns` through untouched. AGENTS.md requires *"API endpoints must validate peer addresses, ASN ranges"*; nothing did.

**Addresses.** The stored value was whatever the client sent. `BgpServer` keys an accepted session by `remoteEndpoint.Address.ToString()`, so any other spelling of the same address produces a peer that can never bind to its session — configured, listed in the UI, and silently never up.

Measured what `IPAddress.TryParse` actually does with the forms a client can send:

| input | parses | `ToString()` | |
|---|---|---|---|
| `1.2.3.4` | ✅ | `1.2.3.4` | canonical |
| `01.02.03.04` | ✅ | `1.2.3.4` | **rewritten** |
| `0x1.2.3.4` | ✅ | `1.2.3.4` | **rewritten** |
| `1.2.3` | ✅ | **`1.2.0.3`** | **a different host than typed** |
| `1.2.3.4 ` | ❌ | — | |
| `::ffff:1.2.3.4` | ✅ | `::ffff:1.2.3.4` | IPv6 family |
| `banana`, `""`, `1.2.3.4.5` | ❌ | — | |

The three-part form is the sharpest: a user typing an incomplete address gets a peer for `1.2.0.3` and no indication anything was reinterpreted.

`NormalizePeerIp` parses, requires `AddressFamily.InterNetwork`, and returns `ToString()` — which collapses every accepted spelling onto the form the BGP path will look for. A missing or unparseable address is now **400**, not a 500 out of a NOT NULL violation.

Note this corrects one detail of the issue: it lists `"1.2.3.4 "` (trailing space) as *"accepted with 200"*. It is accepted by the API today because nothing validates, but `TryParse` rejects it, so the check handles it without special-casing. The forms that genuinely slip through validation-by-TryParse are the rewritten ones above, which is why canonicalizing matters more than rejecting.

**AS numbers.** Exactly four values are rejected:

- `0` — RFC 7607 §2. This one gained urgency from #300: the BGP path now answers an OPEN carrying peer AS 0 with Bad Peer AS, so a peer created with `asn: 0` was accepted, stored, rendered in the UI, and **guaranteed never to establish**, with nothing explaining why. That asymmetry did not exist before the protocol side was tightened.
- `23456` — AS_TRANS is the RFC 6793 placeholder a 4-octet speaker puts in My AS; its real AS arrives in the capability, so no peer ever *is* AS_TRANS.
- `65535` and `4294967295` — the Last ASNs reserved by RFC 7300.

## Departure from the issue: private ranges stay usable

#255 proposes rejecting "optionally 23456 / 65535 / **4200000000+**". That last one is the **RFC 6996 private 32-bit range** (4200000000–4294967294), whose 16-bit counterpart is 64512–65534. Both are exactly what a user of a route server peers with, so excluding them would lock out real peers.

Verified against RFC 7300, which reserves only the two endpoints:

> Operators SHOULD NOT use these Last ASNs for any other purpose or as Private Use ASNs.

— referring to 65535 and 4294967295 specifically, and noting the private range is RFC 6996's, not its own. Tests pin both private ranges as accepted, and the neighbours of every rejected value, so an off-by-one cannot silently lock out a legitimate AS.

## `CustomAsns`

Validated with the same rule on both handlers. Those values are handed to RIPEstat to resolve the prefixes an AS originates, so an unusable one becomes a lookup that can never return anything. The issue is written entirely about the peer's own ASN; this was raised in review of #310 and [recorded on #255](https://github.com/ruhex/BGPLite/issues/255#issuecomment-5440137452) before being fixed here.

## Ordering

All validation runs **before** the store is touched. Combined with #259 making the save a single transaction, a rejected request now leaves nothing behind at all — which is the pair of properties that together make a failed save safe.

Descriptions are capped at 512 characters.

## Verification

`dotnet build BGPLite.sln` + `dotnet test BGPLite.sln`: **726 passed, 0 failed** (693 before, 33 added). `dotnet format --severity warn` clean.

Confirmed to catch the regression — with canonicalization and the ASN filter neutered, 8 of the 33 fail, covering both halves:

```
validation neutered
  Failed NormalizePeerIp_OutputMatchesWhatTheBgpPathWouldStore  x3
  Failed IsConfigurablePeerAsn_RejectsUnusableValues            x4
  Failed NormalizePeerIp_CanonicalizesAcceptedForms(1.2.3 → 1.2.0.3)
```

## Not in scope

Subscription and community name validation is #266's, and the request-body size cap is already handled by `MaxRequestBodyBytes` (#156).

## Summary by CodeRabbit

* **Bug Fixes**
  * Peer creation and updates now reject invalid IP addresses, ASNs, custom ASNs, and overly long descriptions before saving.
  * Peer IP addresses are consistently normalized to standard dotted-quad format.
  * Invalid or reserved ASN values now return a clear bad-request response.
  * Improved consistency between saved peer addresses, BGP sessions, logs, and API responses.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -624,6 +624,11 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
     #region /api/peers
 
+    /// <summary>
+    /// <c>POST /api/peers</c> — creates a peer from the request body. Every field is validated and
+    /// the address canonicalized (#255) BEFORE the store is touched, so a rejected request leaves no
+    /// row behind and an accepted one is stored in the form <c>BgpServer</c> keys sessions by.
+    /// </summary>
     private async Task<ApiResponse> HandleCreatePeer(HttpListenerContext ctx)
     {
         var (body, bodyError) = await ReadBodyAsync(ctx);
@@ -716,6 +721,11 @@ public sealed class ManagementApi : IHostedService, IDisposable
         });
     }
 
+    /// <summary>
+    /// <c>PATCH /api/peers/{id}</c> — updates the supplied fields of an existing peer; an omitted
+    /// collection means "leave it alone", an empty one means "clear it". Validation runs before the
+    /// store is touched, matching <see cref="HandleCreatePeer"/> (#255).
+    /// </summary>
     private async Task<ApiResponse> HandleUpdatePeer(string peerId, HttpListenerContext ctx)
     {
         var peer = _store.GetDbPeerById(peerId);
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -633,6 +633,16 @@ public sealed class ManagementApi : IHostedService, IDisposable
         if (data is null)
             return ApiResponse.Error("Invalid request body", 400);
 
+        // #255: validate everything BEFORE the store is touched. The peer row and its collections
+        // now commit together (#259), so a rejection here leaves nothing behind at all.
+        var normalizedIp = NormalizePeerIp(data.Ip);
+        if (normalizedIp is null)
+            return ApiResponse.Error($"Invalid peer IP: {SanitizeForLog(data.Ip ?? "(missing)")}", 400);
+        if (!IsConfigurablePeerAsn(data.Asn))
+            return ApiResponse.Error($"Invalid peer AS number: {data.Asn}", 400);
+        if (ValidatePeerFields(data.Description, data.CustomAsns) is { } createError)
+            return createError;
+
         var asnLists = data.AsnLists ?? [];
         var customPrefixes = new List<(string Prefix, byte Length)>();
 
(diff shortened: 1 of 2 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

PR #312, opened 2026-08-27
refactor(architecture): require the send path's dependencies instead of defaulting them

### LINKED ISSUE #263: architecture: nullable-optional dependencies across Server/Api/Providers — a missing DI registration silently disables features

ruhex:
Closes #263

## The problem, verified

Every collaborator on the outbound path was a nullable-optional constructor argument threaded `BgpServer` → `BgpSession` → `RouteAssembler`, and each layer supplied a silent fallback for the one below:

```csharp
_prefixAggregator = prefixAggregator ?? new ExactUnionPrefixAggregator();
_communityResolver = communityResolver ?? NullCommunityResolver.Instance;
if (_peerStore is not null && _prefixService is not null && _appConfig is not null) { ... }
```

Dropping one DI registration produced no error at all. Sessions still established; peers just received the shared route table's **seeded** routes instead of the prefixes their operator selected in the UI. That is a wrong route set, not a disabled feature, and #307 could only make it log a warning after the fact.

Worth noting what the old wiring already got wrong without anyone failing: `Program.cs` never passed `prefixAggregator`, so the `?? new ExactUnionPrefixAggregator()` fallback *was* production. The summarization policy for everything on the wire was reachable only as a default argument.

## What changed

**`RouteAssembler`** requires `IPrefixService`, `IPeerStore` and `AppConfig`. The tri-null guard and the shared-table branch under it are gone, along with the `_routeTable` field only that branch used and the null-forgiveness dance around `_appConfig`. It no longer carries a per-session peer label (that is a method parameter now), so it is a plain singleton.

**`SharedTableRouteAssembler`** is the degraded behavior, kept as a named `IRouteAssembler` a caller chooses on purpose. It preserves #289's owner check — the shared table also holds every NLRI peers announced inbound, and serving those to a different peer leaks one tenant's routes — and announces its activation **once** instead of on every refresh.

**`BgpSessionFactory`** owns the session composition. The accept loop needed none of those dependencies for itself, it only forwarded them: `BgpServer` loses five constructor parameters and four fields, and `new BgpSession(…12 args…)` becomes `_sessionFactory.Create(connection, peerConfig)`. The peer-identity callback that `Program.cs` hand-wired as `(ip, asn) => store.UpsertPeer(ip, asn)` is now just `IPeerStore.UpsertPeer` inside the factory.

`BgpSession` drops `_prefixService`, `_appConfig` and `_communityResolver` entirely — it held all three only to hand them to the assembler it built.

**`ManagementApi`** (`IPrefixService`, `IPrefixSourceService`, `ISessionManager`) and **`PrefixService`** (`RipeStatProvider`, `HttpPrefixProvider`) are required too. A null session manager meant a peer edited in the UI was persisted and never pushed to its live session — six `if (_sessionManager is not null)` guards around exactly that. A null http provider meant `GetUserSourcePrefixesAsync` returned `[]`, so a user's URL source silently produced nothing.

**`Program.cs`** registers `IPeerStore`, `IPrefixAggregator`, `IRouteAssembler`, `IBgpSessionFactory` and `TimeProvider` explicitly, and builds the container with `ValidateOnBuild`.

## Acceptance, checked against a running host

> Removing any single DI registration fails loudly at startup

Deleted the `IPrefixAggregator` registration and started the app:

```
Unhandled exception. System.AggregateException: Some services are not able to be constructed
(Error while validating the service descriptor 'ServiceType: BGPLite.Server.IBgpSessionFactory
 Lifetime: Singleton ImplementationType: BGPLite.Server.BgpSessionFactory':
 Unable to resolve service for type 'BGPLite.Routing.IPrefixAggregator'
 while attempting to activate 'BGPLite.Server.BgpSessionFactory'.)
```

Restored, the host starts clean through to `Application started`.

`ValidateScopes` is deliberately **off**, and this is not a judgment call I made blind — with it on the host dies before reaching any of this project's code:

```
Cannot resolve scoped service
'IEnumerable<IDbContextOptionsConfiguration<BgpDbContext>>' from root provider.
   at EntityFrameworkServiceCollectionExtensions.CreateDbContextOptions<TContext>
```

That is EF's own `AddDbContext` plumbing resolving a scoped option-configuration service while constructing the singleton `DbContextOptions`. The app's own singletons already take `IDbContextFactory` rather than a scoped `DbContext`.

> `BgpServer`/`BgpSession` no longer `new` up their collaborators

`BgpServer` still constructs `new SocketBgpConnection(socket)` — that is the transport adapter for the socket it just accepted, has no configuration of its own, and passing a `Socket` into the factory would only couple the seam to `System.Net.Sockets`.

## Tests

`CompositionContractTests` — 17 new tests, 710 total.

- `ProductionDependency_IsRequired` — a `[Theory]` naming each dependency #263 listed (`RouteAssembler.peerStore`, `ManagementApi.sessionManager`, `PrefixService.httpProvider`, …) and asserting it has no default value and `NullabilityState.NotNull`. Re-adding `= null` to any of them fails here rather than silently at the next route send.
- `BgpServer_DoesNotCarryTheSessionsDependencies` — the second acceptance bullet, as a compile-independent assertion.
- Three `SharedTableRouteAssembler` behavior tests: serves the seed but not a route owned by another peer, applies the outgoing filter, warns exactly once across three calls.
- Two factory tests that establish a real session over the `IBgpConnection` seam and assert it asked the **injected** assembler, and that OPEN upserted the peer.

Verified they fail without the fix. Neutering four things independently — `EnumerateUnowned()` → `Enumerate()`, the once-guard, the factory's `routeAssembler:` argument, and `IPeerStore` back to nullable — produces exactly 4 failures out of 17.

`ScriptedConnection` moved out of `BgpSessionRouteOwnershipTests` into a shared fixture so the factory tests can establish a session without a real socket; `ConfigHotReloadTests` gained three inert stubs now that `ManagementApi` cannot be half-wired.

## Not in scope

`PrefixService`'s optional `ILogger`/`TimeSpan?`/`TimeProvider?` parameters are left alone — those are test seams and tuning knobs, not services whose absence changes what a peer receives.

## Summary by CodeRabbit

* **New Features**
  * Added per-peer route assembly for more targeted outbound route announcements.
  * Added a fallback mode that shares only seeded routes while preventing peer-owned routes from being exposed to others.
  * Improved session creation and peer registration consistency.

* **Bug Fixes**
  * Outgoing routes now consistently honor peer-specific filtering.
  * Application startup now detects incomplete configuration immediately instead of failing later.

* **Tests**
  * Added coverage for route isolation, filtering, session setup, and startup composition.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -28,9 +28,12 @@ public sealed class ManagementApi : IHostedService, IDisposable
     private ConcurrencyLimiter? _concurrencyLimiter;
     private IReadOnlyList<string>? _corsAllowedOrigins;
     private readonly BgpMetrics _metrics;
-    private readonly IPrefixService? _prefixService;
-    private readonly IPrefixSourceService? _prefixSources;
-    private readonly ISessionManager? _sessionManager;
+    // #263: required, not optional. Each of these was a silent feature switch: without
+    // _sessionManager a peer edited in the UI was persisted but never pushed to its live session,
+    // and without _prefixService the prefix views reported zero instead of failing.
+    private readonly IPrefixService _prefixService;
+    private readonly IPrefixSourceService _prefixSources;
+    private readonly ISessionManager _sessionManager;
     private readonly ILogger<ManagementApi> _logger;
     private readonly int _port;
     private readonly string _listenAddress;  // #90: bind address — loopback by default
@@ -54,9 +57,9 @@ public sealed class ManagementApi : IHostedService, IDisposable
         AppConfig config,
         BgpMetrics metrics,
         ILogger<ManagementApi> logger,
-        IPrefixService? prefixService = null,
-        IPrefixSourceService? prefixSources = null,
-        ISessionManager? sessionManager = null)
+        IPrefixService prefixService,
+        IPrefixSourceService prefixSources,
+        ISessionManager sessionManager)
     {
         _store = store;
         _routeTable = routeTable;
@@ -597,7 +600,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         // #212: actual advertised count from the live session (post-aggregation, post-dedup).
         var advertisedCount = peer.Asn.HasValue
-            ? _sessionManager?.GetAdvertisedPrefixCount(peer.Ip, peer.Asn.Value) ?? 0
+            ? _sessionManager.GetAdvertisedPrefixCount(peer.Ip, peer.Asn.Value)
             : 0;
 
         return new
@@ -679,8 +682,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         _logger.LogInformation("Created peer {Ip} AS{Asn} ({Id}): {Subs} lists, {Prefixes} custom prefixes, {Asns} custom AS",
             normalizedIp, data.Asn, id, asnLists.Count, customPrefixes.Count, data.CustomAsns?.Count ?? 0);
 
-        if (_sessionManager is not null)
-            _ = _sessionManager.RefreshPeerAsync(normalizedIp, data.Asn);
+        _ = _sessionManager.RefreshPeerAsync(normalizedIp, data.Asn);
 
         return ApiResponse.Ok(new
         {
@@ -771,8 +773,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         _logger.LogInformation("Updated peer {Id}", SanitizeForLog(peerId));
 
-        if (_sessionManager is not null)
-            _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn ?? 0);
+        _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn ?? 0);
 
         return HandleGetPeer(peerId);
     }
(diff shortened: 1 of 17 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

PR #314, opened 2026-08-27
fix(routing): delete a peer's routes when its session ends

### LINKED ISSUE #313: routing: a peer's routes are never removed when its session ends — unbounded growth across reconnects (RFC 4271 §8.2.2)
Verified against `main` @ `3d251d3`.

## Problem

`BgpSession` installs every accepted NLRI into the shared `RouteTable` tagged with itself as owner (#289), but **nothing removes those entries when the session ends**. `RunAsync`'s `finally` sends the Cease, transitions to Idle, updates metrics and the peer's session status — and leaves the routes in place:

```csharp
TransitionTo(BgpFsmState.Idle);
if (wasEstablished)
{
    _metrics.SessionClosed();
    _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false);
}
```

`Dispose` cancels the CTS and disposes the connection; it does not touch the table either. The only removal path in the whole server is `WithdrawIfOwned`, reached exclusively from an explicit WITHDRAWN ROUTES field or a treat-as-withdraw.

## Reproduction (fails on `main`)

Driven through the `IBgpConnection` seam, no sockets:

```csharp
[Fact]
public async Task RoutesAnnouncedByAPeer_AreRemovedWhenItsSessionEnds()
{
    var table = new RouteTable();
    var (session, run, conn) = await EstablishAsync(table);

    conn.EnqueueFrame(AnnounceFrame(8, 0x0A));          // 10.0.0.0/8
    await SettleAsync(conn, () => table.Count == 1);

    session.MarkSilentClose();
    await run.WaitAsync(TimeSpan.FromSeconds(5));
    session.Dispose();

    Assert.Equal(0, table.Count);
}
```

```
Assert.Equal() Failure: Values differ
Expected: 0
Actual:   1
```

The same holds for every teardown reason — the entries outlive a Cease, a hold-timer expiry, a peer NOTIFICATION and a dropped TCP connection alike, because none of those paths touch the table.

## Impact

**Unbounded remote memory growth.** A per-session cap cannot contain this: the peer disconnects, its entries stay, it reconnects with a fresh session and a fresh count, and announces another batch. BGPLite auto-registers unknown peers, so no prior configuration is needed — completing a handshake is enough. This is why it is worth fixing *before* #304: a max-prefix limit that a reconnect resets is not a limit.

**Wrong route count.** `_metrics.SetRouteCount(_routeTable.Count)` and `GET /api/routes` (`_routeTable.GetAll()`) both report the shared table, so both keep counting routes from peers that are long gone.

**Stale data at the prefix.** The entry keeps the dead session as its owner, so it can never be withdrawn by anyone — a later announcement from another peer replaces it, but until then it is immortal.

## RFC

RFC 4271 §8.2.2, Established state, on `TcpConnectionFails`, `NotifMsg`, `HoldTimer_Expires`, `KeepaliveTimer_Expires` failure and every other transition to Idle: the local system

> releases all BGP resources, drops the TCP connection, [and] deletes all routes associated with this connection

§9.1.3 makes the same point from the decision-process side. The exception is RFC 4724 Graceful Restart, where a receiving speaker retains routes as *stale* for the peer's advertised Restart Time and flushes them when it expires — BGPLite does not implement the receiving half at all: `HandleOpen` reads the peer's GR capability and only logs it:

```csharp
var peerGr = CapabilityHelper.GetGracefulRestart(open);
_logger.LogInformation("Peer {Peer} Graceful Restart: {State}", ...);
```

So the current retention is not GR — it is missing cleanup that happens to look like it.

## Recommended fix

Remove the session's owned entries in `RunAsync`'s `finally`, alongside the existing `UpdateSessionStatus`. `RouteTable` already has the ownership tag; it needs a bulk counterpart to `RemoveOwnedBy`:

```csharp
/// <summary>Removes every entry owned by <paramref name="owner"/>; returns how many went.</summary>
public int RemoveAllOwnedBy(object owner)
```

implemented as a scan with the same atomic compare-and-remove `RemoveOwnedBy` uses, so it cannot delete an entry another peer has taken over in the meantime.

Two things to settle in the PR rather than discover in it:

1. **Where to call it.** The `finally` runs for every teardown reason, which is what §8.2.2 wants. It must run even when `wasEstablished` is false — a session that installed nothing removes nothing, so the call is cheap and the guard would only create a hole if the FSM ever changed.
2. **Whether GR should exempt it.** Implementing the receiving half of RFC 4724 is a separate feature (stale marking + Restart Time timer + flush), out of scope here. Until then, flushing unconditionally is the correct, RFC-4271 behavior; retaining without a timer is not GR, it is a leak.

## Test

The reproduction above, plus:
- teardown after a peer NOTIFICATION and after a plain connection close remove the routes too;
- the **startup seed survives** — `RouteSeedingService` writes unowned entries, and a peer's teardown must not take them (the mirror of #307's tenant-isolation property);
- another peer's routes survive;
- a route this session announced but another peer has since replaced is **not** removed by this session's teardown (the same compare-and-remove rule as #289).

## Severity

**HIGH.** Remote, unauthenticated (auto-registration), unbounded memory growth, with no reconnect-resistant mitigation available today. Reported as a precondition for #304 rather than folded into it, since the defect and the fix are independent.

ruhex:
Closes #313

## The defect

RFC 4271 §8.2.2 has a speaker *"delete all routes associated with this connection"* on every transition out of Established. Nothing in BGPLite did. `RunAsync`'s `finally` sent the Cease, transitioned to Idle, updated metrics and the peer's session status — and left the entries in the shared table:

```csharp
TransitionTo(BgpFsmState.Idle);
if (wasEstablished)
{
    _metrics.SessionClosed();
    _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false);
}
```

`WithdrawIfOwned` — reached only from an explicit WITHDRAWN ROUTES field or a treat-as-withdraw — was the **only** removal path in the entire server.

Confirmed on `main` before writing the fix: establish a session, announce `10.0.0.0/8`, tear it down, and `routeTable.Count` is still `1`. Every teardown reason behaves identically, because none of them touches the table.

## Why it is not just tidiness

Because the entries outlive the session, a peer disconnects, reconnects with a fresh session, and adds another batch — indefinitely. BGPLite auto-registers unknown peers, so completing a handshake is the whole prerequisite. That is also why this is filed and fixed **separately from #304** rather than folded into it: a per-session max-prefix cap that a reconnect resets is not a cap, so this had to land first.

`GET /api/routes` (`_routeTable.GetAll()`) and `_metrics.SetRouteCount(_routeTable.Count)` both read the shared table, so both kept counting peers that were long gone.

## The fix

`RouteTable.RemoveAllOwnedBy(object owner)` — the session-close counterpart to `RemoveOwnedBy`, called from `RunAsync`'s `finally`.

Two decisions worth stating rather than leaving to be inferred:

**It uses the same atomic compare-and-remove, per entry.** A bulk clear-by-prefix would delete routes another peer has taken over in the meantime; `AddOrUpdate` transfers ownership, so the losing session must not take the winner's route with it when it closes. Same rule as #289, in bulk form, and there is a test for exactly that.

**It is not gated on `wasEstablished`.** A session that installed nothing removes nothing, so the call is cheap; the guard would only become a hole if the FSM ever grew another way to install routes.

**It is deliberately not GR-exempt.** RFC 4724 lets a receiver retain a restarting peer's routes as *stale* for its advertised Restart Time and flush them when the timer expires. BGPLite implements no part of the receiving half — `HandleOpen` reads the peer's GR capability and only logs it:

```csharp
var peerGr = CapabilityHelper.GetGracefulRestart(open);
_logger.LogInformation("Peer {Peer} Graceful Restart: {State}", ...);
```

Retention without the timer is not Graceful Restart, it is the leak. Implementing the receiving half (stale marking + Restart Time + flush) is a separate feature; until it exists, flushing unconditionally is the RFC 4271 behavior. Recorded in a comment at the call site so the next reader does not have to re-derive it.

## Tests

8 new tests, 701 total.

`BgpSessionRouteOwnershipTests` — the fixture that already owns this machinery, driven through the `IBgpConnection` seam:
- `SessionClose_RemovesTheRoutesThatSessionAnnounced` — the reproduction from the issue.
- `Reconnecting_DoesNotAccumulateTheFormerSessionsRoutes` — the growth-across-reconnects property, which is the actual impact.
- `SessionClose_LeavesTheStartupSeedAndOtherPeersRoutes` — the flush is scoped by owner: the unowned seed and another peer's routes survive. A flush by prefix would empty the table whenever any peer left.
- `SessionClose_DoesNotRemoveARouteAnotherPeerHasSinceReplaced` — the #289 rule at teardown.

`RouteTableTests` — `RemoveAllOwnedBy` in isolation: scoped by owner, skips entries taken over by another owner, an owner that installed nothing removes nothing, null owner throws.

Verified they fail without the fix: replacing `RemoveAllOwnedBy(this)` with `0` turns all four session tests red (7 passed / 4 failed in that fixture).

## Summary by CodeRabbit

* **New Features**
  * Added support for removing all routes associated with a specific session or owner in one operation.

* **Bug Fixes**
  * Routes announced by a session are now automatically withdrawn when the session closes.
  * Cleanup preserves unowned routes, routes from other sessions, and routes whose ownership has changed.
  * Reconnecting sessions no longer accumulate stale routes.

### DIFF
--- a/BGPLite.Routing/RouteTable.cs
+++ b/BGPLite.Routing/RouteTable.cs
@@ -77,6 +77,36 @@ public sealed class RouteTable
             .Remove(new KeyValuePair<(uint Prefix, byte Length), Entry>(key, entry));
     }
 
+    /// <summary>
+    /// Removes every entry still owned by <paramref name="owner"/> and returns how many went — the
+    /// session-close counterpart to <see cref="RemoveOwnedBy"/>. RFC 4271 §8.2.2 has a speaker
+    /// "delete all routes associated with this connection" on every transition out of Established,
+    /// and without it a peer's announcements outlived its session forever: nothing else in the
+    /// server removes an entry, so a peer could reconnect and add another batch indefinitely (#313).
+    /// <para>
+    /// Each removal is the same atomic compare-and-remove <see cref="RemoveOwnedBy"/> performs, so
+    /// an entry another peer has taken over between the scan and the delete stays put — the losing
+    /// session must not delete the winner's route, exactly as in #289. Enumerating a
+    /// <see cref="ConcurrentDictionary{TKey,TValue}"/> while removing from it is safe and defined:
+    /// the enumerator is a moment-in-time view, not a snapshot, and never throws for concurrent
+    /// modification.
+    /// </para>
+    /// </summary>
+    public int RemoveAllOwnedBy(object owner)
+    {
+        ArgumentNullException.ThrowIfNull(owner);
+
+        var removed = 0;
+        foreach (var pair in _routes)
+        {
+            if (!ReferenceEquals(pair.Value.Owner, owner))
+                continue;
+            if (((ICollection<KeyValuePair<(uint Prefix, byte Length), Entry>>)_routes).Remove(pair))
+                removed++;
+        }
+        return removed;
+    }
+
     public Route? Get(uint prefix, byte length) =>
         _routes.TryGetValue((prefix, length), out var entry) ? entry.Route : null;
 
--- a/BGPLite.Server/BgpSession.cs
+++ b/BGPLite.Server/BgpSession.cs
@@ -515,6 +515,26 @@ public sealed class BgpSession : IDisposable
                 catch { /* best-effort */ }
             }
             TransitionTo(BgpFsmState.Idle);
+
+            // RFC 4271 §8.2.2: every transition out of Established "deletes all routes associated
+            // with this connection". #313: nothing did. A peer's announcements outlived its session
+            // forever — no other path in the server removes an entry — so a peer could disconnect,
+            // reconnect and add another batch without limit, and both GET /api/routes and the route
+            // count kept reporting peers that were long gone. Unconditional, not gated on
+            // wasEstablished: a session that installed nothing removes nothing, and the guard would
+            // only be a hole if the FSM ever grew another way to install routes.
+            //
+            // Not GR-exempt. RFC 4724 lets a receiver RETAIN a restarting peer's routes as stale for
+            // its advertised Restart Time and flush them when it expires; BGPLite implements no part
+            // of that (HandleOpen only logs the peer's GR capability), and retention without the
+            // timer is not Graceful Restart, it is the leak.
+            var flushed = _routeTable.RemoveAllOwnedBy(this);
+            if (flushed > 0)
+            {
+                _logger.LogInformation("Removed {Count} route(s) learned from {Peer} on session close", flushed, _peer);
+                _metrics.SetRouteCount(_routeTable.Count);
+            }
+
             if (wasEstablished)
             {
                 _metrics.SessionClosed();
--- a/BGPLite.Tests/BgpSessionRouteOwnershipTests.cs
+++ b/BGPLite.Tests/BgpSessionRouteOwnershipTests.cs
@@ -282,6 +282,108 @@ public class BgpSessionRouteOwnershipTests
         return Frame(BgpMessageType.Update, [.. payload]);
     }
 
+    // ---- #313: a session's routes go when the session goes ----
+
+    /// <summary>
+    /// RFC 4271 §8.2.2: every transition out of Established "deletes all routes associated with this
+    /// connection". #313: nothing did. A peer's announcements outlived its session, and since no
+    /// other path in the server removes an entry, a peer could disconnect, reconnect and add another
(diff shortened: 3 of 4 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

PR #315, opened 2026-08-27
perf(routing): memoize community normalization per Aggregate call

### LINKED ISSUE #305: perf(routing): AttributeKey.From allocates 4x per route on every send — ~240k allocations for a 60k-route dump
Split out of #292 (item 8). Verified against `main` @ 9708cdc.

## Problem

`ExactUnionPrefixAggregator.AttributeKey.From(route)` runs **once per route**, and allocates four times per call:

```csharp
private static uint[] NormalizeCommunities(IReadOnlyList<uint> communities)
{
    var sorted = communities.Distinct().ToArray();   // Enumerable.Distinct allocates an internal Set<uint>
    Array.Sort(sorted);
    return sorted;
}
```

`Distinct()` allocates a set, `ToArray()` allocates the array — twice over, for regular and large communities.

## Execution path and impact

`Aggregate` is the choke point for **both** the initial route dump and every `RefreshRoutesAsync`:

`BgpSession.SendRoutesAsync` → `_prefixAggregator.Aggregate(routes)` → `AttributeKey.From` per route.

For a 60 000-prefix send that is roughly 240 000 allocations before a single byte reaches the wire, repeated per peer per refresh. `PrefixAutoRefreshService` and the `onSourceChanged` convergence callback fan this out across all established sessions concurrently via `RefreshAllEstablishedAsync`, so the peaks coincide.

## Why it is avoidable

The inputs make almost all of that work redundant. `RouteAssembler.MakeRoute` passes the **same** `comms` array instance to every route built from one source:

```csharp
var comms = _communityResolver.Resolve(new CommunitySource(CommunitySourceKind.AsnList, list.Name));
var prefixes = await prefixService.GetPrefixesForAsns(list.Asns, ct);
foreach (var (prefix, length, _) in prefixes)
    routes.Add(MakeRoute(prefix, length, nextHop, null, comms));
```

So a typical send has a handful of distinct community-set instances across tens of thousands of routes, and normalizes each one thousands of times.

## Recommended fix

1. Short-circuit `Count <= 1` — no dedup or sort is needed, and this is the overwhelmingly common case (most routes carry zero or one community).
2. Add a reference-equality fast path: memoize the normalized key per distinct backing-array instance for the duration of one `Aggregate` call (a `Dictionary<object, uint[]>` with reference comparison, or just cache the last-seen instance since routes arrive grouped by source).

Both are local to `AttributeKey`; no public API changes.

## Acceptance

A benchmark or allocation measurement **before and after** on a representative send (say 60 000 routes across 3 community sets) attached to the PR. This is a performance claim, so it should not merge on reasoning alone — if the measurement does not show a meaningful difference, close the issue rather than taking the change.

## Severity

**LOW (performance).** No correctness impact; the cost is GC pressure on the advertise path, concentrated exactly when many peers refresh at once.

ruhex:
Closes #305

The issue asks for a measurement rather than an argument, and to close it unfixed if the numbers do not justify the change. They do.

## What was happening

`AttributeKey.From` ran once per route and allocated four times — `Distinct()` builds an internal `Set<uint>`, `ToArray()` builds the array, twice over for regular and large communities. `Aggregate` is the choke point for both the initial dump and every `RefreshRoutesAsync`, and `PrefixAutoRefreshService` / the source-change convergence callback fan it out across all established sessions at once, so the peaks coincide.

Nearly all of that work was redundant. `RouteAssembler.MakeRoute` passes the **same** resolved community array instance to every route built from one source:

```csharp
var comms = _communityResolver.Resolve(new CommunitySource(CommunitySourceKind.AsnList, list.Name));
foreach (var (prefix, length, _) in prefixes)
    routes.Add(MakeRoute(prefix, length, nextHop, null, comms));
```

So a 60 000-route send normalizes a handful of distinct instances tens of thousands of times.

## The change

`KeyNormalizer` memoizes each normalized set by the **identity** of the collection it came from, for the duration of a single `Aggregate` call. It is a mutable struct held as a local with lazily-created dictionaries, so a send whose routes carry no communities — the seeded shared table — allocates nothing for it, and the empty case short-circuits before `Distinct()` runs at all.

Keying by reference is not a correctness shortcut. `AttributeKey.Equals` stays structural, so two distinct instances holding equal content normalize twice and still group together; identity is only used to skip repeat work, never to decide equivalence. There is a test for exactly that.

## Measurement

`GC.GetAllocatedBytesForCurrentThread()` around `ExactUnionPrefixAggregator.Aggregate`, Release build, 20–2000 iterations per scenario after a 3-iteration warm-up, allocation and wall time **per call**. Routes are built the way the send path builds them: one shared community array per source, interleaved across sources (the pessimistic ordering for any last-seen cache).

| scenario | alloc before | alloc after | | ms before | ms after | |
|---|---:|---:|---|---:|---:|---|
| 60k routes / 3 sets / 2 comms + 1 large | 47,286,152 B | 8,911,660 B | **5.3×** | 18.873 | 10.931 | **1.7×** |
| 60k routes / 3 sets / 1 comm | 26,668,532 B | 8,909,933 B | **3.0×** | 10.995 | 5.290 | **2.1×** |
| 60k routes / 3 sets / no communities | 3,450,224 B | 3,450,631 B | flat | 2.674 | 1.846 | 1.4× |
| 11k routes / 1 set / 1 comm | 3,959,633 B | 704,379 B | **5.6×** | 0.849 | 0.440 | 1.9× |
| 200 routes / 1 set / 2 comms + 1 large | 141,072 B | 14,368 B | **9.8×** | 0.021 | 0.008 | 2.6× |

The no-communities row is the case the memo cannot help; it is flat on allocation (+407 B is run-to-run noise) and still faster, because `Distinct()` allocated a set even for an empty input and the short-circuit does not. Nothing regresses, including the 200-route send where the per-call overhead of a normalizer would show up if it existed.

## Tests

3 new tests in `PrefixAggregatorTests`, 696 total, pinning the properties the memo could plausibly break:

- `EqualCommunitiesInDistinctInstances_StillGroupTogether` — equal content in different array instances, in different orders, still merges into one group. This is the assertion that would fail if identity were ever mistaken for equivalence.
- `SharedCommunityInstances_GroupPerInstanceContent` — the shape the memo is built for: one array shared across eight routes from each of two sources, which must stay two groups.
- `LargeCommunities_MemoizedSeparatelyFromRegularOnes` — the two memos are independent, and routes with and without large communities separate correctly.

Writing the third one surfaced something worth stating: the emitted route carries the **group template's own** community arrays, not the normalized ones, so sorting affects grouping only and never what goes on the wire. That was already true before this PR; the test now says so.

The benchmark harness is not committed — it references `BGPLite.Routing` from a scratch console project and would only be dead weight in the repo. The numbers above are reproducible from the scenario table in the test file's shape.

### DIFF
--- a/BGPLite.Routing/ExactUnionPrefixAggregator.cs
+++ b/BGPLite.Routing/ExactUnionPrefixAggregator.cs
@@ -36,9 +36,16 @@ public sealed class ExactUnionPrefixAggregator : IPrefixAggregator
         // Capacity is the expected number of DISTINCT community sets, not route count.
         // A typical send carries 1-5 community sets even with tens of thousands of routes.
         var groups = new Dictionary<AttributeKey, List<Route>>(4);
+        // #305: normalization was per route — Distinct().ToArray() twice over, so four allocations
+        // for every route on every send. RouteAssembler hands every route built from one source the
+        // SAME community array instance, so a 60k-route dump normalizes a handful of distinct
+        // instances tens of thousands of times. The normalizer memoizes by instance for the duration
+        // of this call; it is a struct with lazily-created dictionaries, so a send whose routes carry
+        // no communities allocates nothing for it at all.
+        var normalizer = default(KeyNormalizer);
         foreach (var route in source)
         {
-            var key = AttributeKey.From(route);
+            var key = normalizer.KeyFor(route);
             if (!groups.TryGetValue(key, out var group))
             {
                 group = new List<Route>();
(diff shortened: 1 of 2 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

Issue #290, opened 2026-08-26
protocol: attribute flags never validated against the type code, ORIGIN/NEXT_HOP lengths only lower-bounded — no Attribute Flags Error (4) / Attribute Length Error (5) on inbound (RFC 4271 §6.3)


ruhex:
## Summary

`ParseRouteAttributes` validates attribute *values* but never validates the attribute **flags against the type code**, and it lower-bounds rather than fixes the lengths of the two fixed-width well-known attributes. RFC 4271 §6.3 requires:

> If any recognized attribute has Attribute Flags that conflict with the Attribute Type Code, then the Error Subcode MUST be set to Attribute Flags Error.
> If any recognized attribute has an Attribute Length that conflicts with the expected length (based on the attribute type code), then the Error Subcode MUST be set to Attribute Length Error.

`BgpConstants.SubError.AttributeFlagsError` (4) and `AttributeLengthError` (5) already exist (#235); nothing in the inbound pipeline raises them for these cases. #272 added the reserved-bit 0x08 check, which is the *only* flag validation present.

## Affected code

`BGPLite.Protocol/UpdateCodec.cs` — `ParseRouteAttributes`:

```csharp
case BgpConstants.Attribute.Origin:
    if (attr.Data.Length < 1)                       // RFC 4271 §4.3: ORIGIN length is exactly 1
        throw new BgpNotificationException(..., InvalidOriginAttribute, "Malformed ORIGIN attribute");
    AttributeHelper.ReadOrigin(attr);
    originSeen = true;
    break;
...
case BgpConstants.Attribute.NextHop:
    if (attr.Data.Length < 4)                       // RFC 4271 §4.3: NEXT_HOP length is exactly 4
        throw new BgpNotificationException(..., InvalidNextHopAttribute, "Malformed NEXT_HOP attribute");
    nextHop = AttributeHelper.ReadNextHop(attr);
    nextHopSeen = true;
    break;
```

`PathAttribute` exposes `Optional` / `Transitive` / `Partial` but nothing reads them on the inbound path.

## Reproduction

All of the following are accepted, and the route is installed:

**Flags conflicting with the type code** (expected: Update Message Error / subcode 4):

| Attribute | Flags on the wire | Correct flags | Result |
|---|---|---|---|
| ORIGIN | `0x80` (Optional) | `0x40` well-known transitive | ACCEPTED |
[...]

Issue #287, opened 2026-08-26
protocol: duplicate path attributes silently accepted, last occurrence wins — NEXT_HOP/COMMUNITY smuggling (RFC 4271 §6.3, RFC 7606 §3(g))


ruhex:
## Summary

`UpdateCodec.ParseRouteAttributes` walks `update.PathAttributes` with a plain `switch` and assigns unconditionally, so a repeated attribute simply overwrites the previous one — **last occurrence wins**. RFC 4271 §6.3 requires the opposite:

> If any attribute appears more than once in the UPDATE message, then the Error Subcode MUST be set to Malformed Attribute List.

RFC 7606 §3(g) softens the action to *treat-as-withdraw* but keeps the rejection: the UPDATE must not be applied.

## Impact

**Attribute smuggling.** An UPDATE carrying two `NEXT_HOP` attributes installs the *second* one. Anything that inspects the first occurrence — a monitoring collector, a looking-glass, a policy middlebox, an operator reading a packet capture — sees a different next hop than the one BGPLite actually installs in the `RouteTable`. The same divergence applies to `ORIGIN`, `AS_PATH`, `COMMUNITY`, `LARGE_COMMUNITY` and `AGGREGATOR`; duplicating `COMMUNITY` is a direct way to defeat `PeerCommunityFilter`'s allow-set (the first, benign community set is what a reviewer sees; the second is what is stored).

The session stays up and nothing is logged — `metrics.UpdatesRejected` stays at 0.

## Affected code

`BGPLite.Protocol/UpdateCodec.cs` — `ParseRouteAttributes`:

```csharp
foreach (var attr in update.PathAttributes)
{
    switch (attr.TypeCode)
    {
        case BgpConstants.Attribute.Origin:
            ...
            originSeen = true;                       // set again, no duplicate check
            break;
        case BgpConstants.Attribute.NextHop:
            ...
            nextHop = AttributeHelper.ReadNextHop(attr);   // overwrites silently
            nextHopSeen = true;
            break;
        ...
    }
}
```

The `*Seen` flags exist only to feed `ValidateMandatoryAttributes` (presence), never to detect a second occurrence.

## Reproduction

Codec level:

```
UPDATE attrs = ORIGIN(igp) ORIGIN(incomplete) AS_PATH[100] NEXT_HOP 192.0.2.1, NLRI 10.0.0.0/8
  -> reader accepted 4 attributes
[...]

PR #333, opened 2026-08-27
fix(protocol): reject an unrecognized well-known path attribute with subcode 2 (#322)

### LINKED ISSUE #322: protocol: an unrecognized well-known path attribute is silently accepted — routes install with ignored semantics (RFC 4271 §6.3 subcode 2)

## Problem

`UpdateCodec.ExpectedAttributeShape` returns `null` for any type code it does not know (`UpdateCodec.cs:336`), `ValidateAttributeShape` early-returns on that (`:346-347`), and the switch in `ParseRouteAttributes` (`:254-288`) has no default branch. Net effect: an UPDATE carrying a **well-known** attribute (Optional bit = 0) of an unknown type is fully accepted and its NLRI installed — the attribute's semantics are silently ignored.

```csharp
_ => null,                                   // UpdateCodec.cs:336
...
if (ExpectedAttributeShape(attr.TypeCode, fourByteAsnSession) is not { } expected)
    return;                                  // UpdateCodec.cs:346-347
```

RFC 4271 §6.3 requires the "Unrecognized Well-known Attribute" error (Update Message Error, subcode 2 — originally a session reset); RFC 7606 revises the remedy for attribute errors down to treat-as-withdraw. Neither is implemented. Note the constant `BgpConstants.SubError.UnrecognizedWellKnownAttribute` (`BgpConstants.cs:57`) exists and is **never used** — the check was planned, not landed.

The docstring on `ExpectedAttributeShape` (`:306-316`) documents a deliberate scope decision (don't validate attributes the codec doesn't consume, to avoid over-rejecting — the #290 lesson). That reasoning covers *optional* attributes fine, but it does not cover unknown **well-known** attributes: no RFC permits ignoring those. The deviation is not in `docs/DESIGN_DECISIONS.md`.

Scope of the fix stays narrow — only reject the well-known case:
- unknown **optional transitive** → keep ignoring (correct today, RFC 4271 §6.3 propagate-and-partial semantics; BGPLite re-originates so nothing to propagate);
- unknown **optional non-transitive** → discard attribute, continue (RFC 7606 §2; current ignore is materially equivalent for this codebase);
- unknown **well-known (Optional=0)** → reject, subcode 2, treat-as-withdraw.

## Execution path and impact

Peer sends UPDATE with NLRI + attribute type 99, flags 0x40 → wire parsing succeeds → `ParseRouteAttributes` skips the attribute → route installed in the shared `RouteTable`. A buggy or experimental peer implementation gets its mandatory-semantics attribute (present or future type code) ignored without any signal. For a route server this is the "accept and re-advertise with ignored policy semantics" class — silent, no log, no NOTIFICATION.

## Fix

In `ParseRouteAttributes`, next to the duplicate guard / `ValidateAttributeShape` call:

```csharp
if (!attr.Optional && !AttributeHelper.IsKnownAttribute(attr.TypeCode))
    throw new BgpNotificationException(
        BgpConstants.Error.UpdateMessageError,
        BgpConstants.SubError.UnrecognizedWellKnownAttribute,
        $"Unrecognized well-known attribute type {attr.TypeCode}");
```

It lands in the existing treat-as-withdraw catch unchanged. Alternatively, if silent acceptance is a deliberate choice, record it in `docs/DESIGN_DECISIONS.md` — today it is documented only in a code comment.

## Test

- Unit: UPDATE with NLRI + attribute type 99, flags 0x40 → `ParseRouteAttributes` throws 3/2.
- Wire-level (scripted connection): same UPDATE → session survives, NLRI withdrawn (treat-as-withdraw), exactly no NOTIFICATION beyond the standard path.
- Negative control: unknown **optional** attribute (flags 0xC0, type 99) → still accepted (behavior unchanged).

ruhex:
Closes #322

## Problem

`ParseRouteAttributes` silently skipped any type code it did not know: `ExpectedAttributeShape` returned `null` (`UpdateCodec.cs` `_ => null`), `ValidateAttributeShape` early-returned, and the attribute switch had no default branch. An UPDATE carrying a **well-known** attribute (Optional bit = 0) of an unknown type installed its NLRI with the attribute's semantics ignored — no log, no NOTIFICATION. RFC 4271 §6.3 mandates the Unrecognized Well-known Attribute error (subcode 2); RFC 7606 lowers the remedy to treat-as-withdraw. `BgpConstants.SubError.UnrecognizedWellKnownAttribute` was declared but never used — the check was planned, not implemented.

## Fix

One guard in `ParseRouteAttributes`, after the #287 duplicate guard and before #290 shape validation:

```csharp
if (!attr.Optional && !AttributeHelper.IsKnownAttribute(attr.TypeCode))
    throw new BgpNotificationException(UpdateMessageError, UnrecognizedWellKnownAttribute, ...);
```

It lands in the **existing** RFC 7606 treat-as-withdraw path (same as Missing Well-known / Attribute Flags Error): the UPDATE's NLRI is withdrawn, `UpdatesRejected` increments, the session stays up, no NOTIFICATION.

Deliberate scope split (the over-rejection lesson of #290):
- unrecognized **optional** attribute → still ignored (RFC 7606 §2);
- known-but-unread well-known attributes (**LOCAL_PREF, ATOMIC_AGGREGATE**; MED is optional) → still accepted — guarded by a dedicated test;
- the #290 fixture `ParseRouteAttributes_UnrecognizedAttribute_IsNotShapeChecked` used `Flags = 0x00` (i.e. well-known) for its unknown attribute — updated to optional flags so it guards the surviving #290 invariant, with the doc comment linking to the new well-known test.

## Verification

- `dotnet build BGPLite.sln` — 0 errors; `dotnet test BGPLite.sln` — **764/764**.
- Red on old code is proven by the #290 fixture itself: before this change an UPDATE with `Flags=0x00, TypeCode=200` parsed successfully (that fixture failed red the moment the guard landed).
- New tests: direct `ParseRouteAttributes` → 3/2; unrecognized optional still parses; LOCAL_PREF + ATOMIC_AGGREGATE still accepted; wire-level (`MalformedUpdateResilienceTests`) — established session, good announcement installed, then the same NLRI re-announced with a type-99/0x40 attribute → prefix withdrawn, session Established, `UpdatesRejected == 1`, zero NOTIFICATIONs on the wire.
- Reviewer: PRISTINE (one MINOR — stale subcode list in the method's XML doc — fixed before the push).

## Notes

- Duplicate unknown well-known attribute: the later occurrence is discarded by the #287 guard before this check — consistent with RFC 7606 §3 (discarded ≠ validated); the first occurrence rejects.
- MP_REACH/UNREACH (14/15), EXTENDED_COMMUNITIES (16) etc. are optional by their RFCs, so the guard cannot fire on them unless the peer also missets the Optional flag — in which case rejection is correct (`IsKnownAttribute` covers every IANA well-known type: 1,2,3,5,6).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unrecognized BGP path attributes according to RFC 4271.
  * Unrecognized mandatory attributes are now rejected with the appropriate parse error.
  * Unrecognized optional attributes continue to be ignored safely.
  * Known attributes that are not fully processed remain accepted.
  * Malformed updates are withdrawn without terminating the established session or sending a notification.

### DIFF
--- a/BGPLite.Protocol/UpdateCodec.cs
+++ b/BGPLite.Protocol/UpdateCodec.cs
@@ -200,7 +200,7 @@ public static class UpdateCodec
     /// mandatory well-known set (ORIGIN/AS_PATH/NEXT_HOP), AS4_PATH trailing-sequence
     /// reconstruction, and AGGREGATOR/AS4_AGGREGATOR consistency. On a 4-octet session
     /// (<paramref name="fourByteAsnSession"/>) AS4_PATH/AS4_AGGREGATOR are skipped per RFC 6793 §4.1.
-    /// Throws <see cref="BgpNotificationException"/> carrying the RFC subcode (3/6/8/9/11) so the
+    /// Throws <see cref="BgpNotificationException"/> carrying the RFC subcode (2/3/4/5/6/8/9/11) so the
     /// caller can apply treat-as-withdraw (RFC 7606) — the exact pipeline previously inlined in
     /// BgpSession.HandleUpdateAsync, moved verbatim (#270).
     /// </summary>
@@ -249,6 +249,20 @@ public static class UpdateCodec
                 // never shape-checked. RFC 7606 §3 says those occurrences are discarded, not
                 // "discarded but still validated" — checking them would reject an UPDATE over an
                 // attribute that has no effect on the result (#287 + #290).
+                //
+                // RFC 4271 §6.3 (#322): an attribute with the Optional bit clear that this codec
+                // does not recognize has unknown well-known semantics — it MUST be rejected with
+                // Unrecognized Well-known Attribute (subcode 2), never silently ignored. Only
+                // unrecognized OPTIONAL attributes may be discarded (RFC 7606 §2), and those fall
+                // through untouched here. The throw routes through the caller's treat-as-withdraw,
+                // not a session reset. Known-but-unread attributes (LOCAL_PREF,
+                // ATOMIC_AGGREGATE; MED is optional) pass IsKnownAttribute and stay accepted —
+                // over-rejecting those was the #290 lesson.
+                if (!attr.Optional && !AttributeHelper.IsKnownAttribute(attr.TypeCode))
+                    throw new BgpNotificationException(
+                        BgpConstants.Error.UpdateMessageError,
+                        BgpConstants.SubError.UnrecognizedWellKnownAttribute,
+                        $"Unrecognized well-known path attribute type {attr.TypeCode}");
                 ValidateAttributeShape(attr, fourByteAsnSession);
 
                 switch (attr.TypeCode)
--- a/BGPLite.Tests/BgpSessionRfc6793Tests.cs
+++ b/BGPLite.Tests/BgpSessionRfc6793Tests.cs
@@ -107,6 +107,75 @@ public class BgpSessionRfc6793Tests
     }
 
     [Fact]
+    public void ParseRouteAttributes_UnrecognizedWellKnownAttribute_ThrowsSubcode2()
+    {
+        // RFC 4271 §6.3 / #322: an Optional=0 attribute of an unknown type code must be rejected
+        // (subcode 2) — routes must not install with unknown well-known semantics attached.
+        var update = new BgpUpdateMessage
+        {
+            PathAttributes =
+            [
+                AttributeHelper.WriteOrigin(BgpOrigin.Igp),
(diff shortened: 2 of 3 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

Issue #262, opened 2026-08-26
architecture: PeerStore/IPeerStore is fully synchronous on async paths — thread starvation under contention; move the contract to async (refs #10)


ruhex:
## What is wrong

The entire persistence contract is synchronous: `IPeerStore.cs:5-19,31` (all `void`/`List<>` methods, including `LoadPeerRoutingView`), implemented with sync EF calls (`PeerStore.cs:171-199`, `ToList`/`SaveChanges`/`ExecuteUpdate`). It is invoked from async contexts: the BGP send path (`RouteAssembler.cs:75,251`), session teardown/status writes (`BgpSession.cs:406,1167`), and every management-API handler (`ManagementApi.cs:578-913`). `SqlitePragmasInterceptor.cs:25` sets `busy_timeout = 5s`, so each contending writer blocks a thread-pool thread up to 5 s; with the 64-slot in-flight cap that is up to 64 blocked threads. This also contradicts AGENTS.md: "Prefer async I/O for database and HTTP operations".

## Impact

Under write contention (peers mutating via API while sessions establish/refresh) the thread pool starves; keepalive loops of unrelated sessions slow down. Every month the sync signature survives, the eventual migration gets more expensive (interface + implementation + all callers).

## Fix

Expand–Contract:
1. Add async twins to `IPeerStore` (`LoadPeerRoutingViewAsync`, `UpsertPeerAsync`, `UpdateSessionStatusAsync`, …).
2. Migrate callers (RouteAssembler first — the only hot-path call; then ManagementApi handlers).
3. Remove the sync members.

Raw ADO.NET upserts move to `ExecuteScalarAsync`. Should land together with (or after) the Contracts extraction planned on #230 — the interface is the thing being moved.

## Acceptance

- No sync DB I/O remains on async paths (grep `peerStore.` / `_store.` in async methods).
- `LoadPeerRoutingViewAsync` used by RouteAssembler; measured establish/refresh latency under concurrent API writes does not degrade.

Refs: audit 2026-08-26; #10 (P6 epic — "async PeerStore" item), #230.

PR #334, opened 2026-08-27
fix(server): guard UpdateSessionStatus in the RunAsync finally (#325)

### LINKED ISSUE #325: server: a failing UpdateSessionStatus in the RunAsync finally faults the session task and leaks PeerCount/DB status forever

## Problem

The teardown `finally` of `BgpSession.RunAsync` (`BgpSession.cs:533-539`) makes a synchronous EF Core call with no guard, **between** two metric updates:

```csharp
if (wasEstablished)
{
    _metrics.SessionClosed();
    _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false);   // sync DB UPDATE, unguarded
}
_metrics.PeerDisconnected();        // skipped if the line above throws
_logger.LogInformation("SessionClosed with {Peer}", _peer);
```

`UpdateSessionStatus` is `ExecuteSqlInterpolated` on SQLite (`PeerStore.cs:96-115`): it throws `SqliteException` on `database is locked` past `busy_timeout` / disk I/O errors. Concurrent writers to the same DB exist by design: `UpsertPeer` on every peer handshake, the management API, auto-refresh (the sync-store shape itself is #262 — this issue is only about the unguarded call site).

When it throws:

1. The exception replaces the normal completion of `RunAsync`.
2. `BgpServer.RunSessionAsync` (`BgpServer.cs:256-274`) is `try/finally` **without a catch** and was started fire-and-forget (`_ = RunSessionAsync(...)`, `BgpServer.cs:244`) → a faulted, unobserved task. No `TaskScheduler.UnobservedTaskException` handler exists, so the exception disappears without a single log line — violating the repo rule that every best-effort catch either logs or documents its silence (here nobody even catches).
3. `_metrics.PeerDisconnected()` is skipped although `PeerConnected()` ran at session start (`BgpSession.cs:275`) → `BgpMetrics.PeerCount` drifts **up permanently** and is exposed through the UI/API (ghost peers).
4. The DB row keeps `Status = active` forever.

## Counter-evidence checked

All shutdown tests (`BgpSessionShutdownTests`) use no-op/fake stores whose `UpdateSessionStatus` does nothing; no test covers a throwing store. `PeerStoreAtomicityTests` cover API-side atomicity, not this call site. Every other external call in the same finally is already guarded (`SendNotificationAsync` earlier in teardown) or non-throwing by construction.

## Fix

Wrap the call in the same best-effort pattern the block already uses elsewhere:

```csharp
try { _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false); }
catch (Exception ex) { _logger.LogWarning(ex, "Failed to persist session status for {Peer}", _peer); }
```

so `PeerDisconnected()` and the closing log always run and `RunAsync` always completes normally.

## Test

Established session over a fake `IBgpConnection` + an `IPeerStore` whose `UpdateSessionStatus` throws → assert: `RunAsync` completes (not faulted), `metrics.PeerCount` returns to the pre-session value, teardown NOTIFICATION/flush behavior unchanged.

ruhex:
Closes #325

## Problem

`RunAsync`'s teardown finally called the synchronous `_peerStore?.UpdateSessionStatus(...)` unguarded, **between** two metric updates. On a transient store failure (SQLite `database is locked` past `busy_timeout`; concurrent writers exist by design — `UpsertPeer` per handshake, the management API, auto-refresh) the exception escaped `RunAsync`. `RunSessionAsync` (`BgpServer`) is `try/finally` without a catch and is started fire-and-forget — the task faulted **unobserved**:

- the exception disappeared without a single log line (violating the repo rule that every best-effort catch either logs or documents silence — here nobody even caught);
- `PeerDisconnected()` was skipped although `PeerConnected()` had run → `BgpMetrics.PeerCount` drifted up permanently (ghost peers in the UI/API);
- the peer's DB row kept `Status = active` forever.

## Fix

The call is wrapped in the same best-effort pattern the block already uses for `SendNotificationAsync` — `try/catch` + `LogWarning` — so `PeerDisconnected()` and the closing log always run and `RunAsync` always completes normally. The sync store contract itself remains #262; this PR only contains the call site.

## Verification

- `dotnet test` — **765/765**.
- New regression test `RunAsync_StoreFailureInFinally_DoesNotFaultTask_AndDecrementsPeerCount`: an established scripted session with a store whose `UpdateSessionStatus` throws → `RunAsync` completes, `PeerCount` returns to baseline. Verified **red** on the unguarded code (temporarily reverted: the store exception rethrown out of `WaitAsync`, exactly the production failure mode).
- Internal reviewer: PRISTINE (sync `UpdateSessionStatus` cannot produce caller-cancellation OCE, so the no-OCE-swallowing policy is not affected).

### DIFF
--- a/BGPLite.Server/BgpSession.cs
+++ b/BGPLite.Server/BgpSession.cs
@@ -533,7 +533,13 @@ public sealed class BgpSession : IDisposable
             if (wasEstablished)
             {
                 _metrics.SessionClosed();
-                _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false);
+                // Best-effort, like SendNotificationAsync above: this sync DB write runs in the
+                // finally of a fire-and-forget task (RunSessionAsync has no catch), so a transient
+                // store failure (SQLite "database is locked" past busy_timeout) must not escape —
+                // it would fault RunAsync unobserved, skip PeerDisconnected() below, and leak
+                // PeerCount plus the row's Status=active forever (#325).
+                try { _peerStore?.UpdateSessionStatus(_peerConfig.Address, _remoteAsn, false); }
+                catch (Exception ex) { _logger.LogWarning(ex, "Failed to persist session status for {Peer}", _peer); }
             }
             _metrics.PeerDisconnected();
             _logger.LogInformation("SessionClosed with {Peer}", _peer);
--- a/BGPLite.Tests/BgpSessionShutdownTests.cs
+++ b/BGPLite.Tests/BgpSessionShutdownTests.cs
@@ -73,6 +73,74 @@ public class BgpSessionShutdownTests
         Assert.Equal(BgpConstants.SubError.CeaseAdministrativeReset, notif.SubErrorCode);
     }
 
+    /// <summary>
+    /// Regression #325: UpdateSessionStatus in the RunAsync finally is a best-effort DB write on a
+    /// fire-and-forget task (RunSessionAsync has no catch) — a transient store failure must not
+    /// fault RunAsync, or the exception vanishes unobserved, PeerDisconnected() is skipped, and
+    /// PeerCount (plus the peer's Status=active row) leaks forever.
+    /// </summary>
+    [Fact]
+    public async Task RunAsync_StoreFailureInFinally_DoesNotFaultTask_AndDecrementsPeerCount()
+    {
+        var metrics = new BgpMetrics();
+        var conn = new ScriptedConnection();
+        using var session = new BgpSession(
+            conn,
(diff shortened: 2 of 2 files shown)


ruhex:
@coderabbitai review

ruhex:
Merged without a CodeRabbit pass: the free-OSS review quota was exhausted at PR creation ("next included review in ~44 min") and the maintainer opted not to wait. Compensating gates: green CI (build-test / format-check / analyze / CodeQL), a dedicated internal adversarial review (PRISTINE), and a red-verified regression test. If CodeRabbit findings arrive after the quota resets, they will be handled as follow-ups.

Issue #327, opened 2026-08-27
config: a malformed PrefixSources Community aborts the whole seeding loop and Validate() never checks PrefixSources


ruhex:
Verified against `main` @ `26d474e`.

## Problem

Two halves of one gap.

**1. Seeding aborts on the first bad community.** `RouteSeedingService.SeedAsync` parses the tag with a bare `CommunityCodec.Parse(source.Community!)` inside the per-source loop (`RouteSeedingService.cs:66-68`). A non-numeric value (`Community: "65000:abc"`) throws `FormatException`; the outer catch-all (`:98-103`) then aborts the entire `SeedAsync` — sources **after** the broken one are never seeded, and `WarmUpAsync` plus the final `RefreshAllEstablishedAsync` (`:88-92`) are skipped. The server starts its listeners and quietly serves a partial table (degraded `/api/routes`, cold RIPEstat cache) with only a single `LogError` line. Contrast: everywhere else the same input is deliberately non-throwing — `ConfigCommunityResolver` logs-and-falls-back ("never throw during a peer send").

**2. `Validate()` ignores `PrefixSources`.** `AppConfig.Validate()` (`AppConfig.cs:99-127`) checks `Bgp`, `ApiPort`, `MaxRequestBodyBytes`, and `Peers[*].Address` — nothing about `PrefixSources` (no `Community` format, no `Kind`, no required `Url`/`Path`/`Asn` per kind). This contradicts the fail-loud-at-startup policy (#89 / #8): the operator's typo is discovered *after* the host is up instead of at config load.

## Execution path and impact

`PrefixSources: [{Kind: file, Name: a, Path: nets.txt, Community: "65000:abc"}, {Kind: file, Name: b, Path: more.txt}]` → startup passes validation → listeners start → seeding loads source `a`'s prefixes, throws on its community, source `b` is never seeded, warm-up never runs, established-at-startup peers never get the final push. Partial table until restart.

(Note: a community *value* over 65535 does **not** throw today — it silently truncates; that's a separate codec bug.)

## Fix

1. Per-source resilience in `SeedAsync`: try/catch around the community parse with a warning and an untagged fallback (or reuse `ConfigCommunityResolver`'s ParseOrDefault semantics) so one bad tag degrades that source only.
2. Validate `PrefixSources` in `AppConfig.Validate()` (refs epic #8): `Community` via `CommunityCodec`, `Kind` in {file, http, asn}, required fields per kind, so bad config fails before the host is built.

## Test

[...]

PR #335, opened 2026-08-27
fix(protocol): reject a community VALUE above 65535 instead of masking it (#328)

### LINKED ISSUE #328: protocol: CommunityCodec.Parse silently truncates a community VALUE above 65535 instead of rejecting it

## Problem

`CommunityCodec.Parse` (`CommunityCodec.cs:9-22`) validates the ASN half but silently masks the VALUE half:

```csharp
if (asn > 0xFFFF)
    throw new FormatException($"... ASN part must be 0-65535 (got {asn}).");
return (asn << 16) | (value & 0xFFFF);      // "65000:70000" -> 65000:4464, no error
```

The two halves of the same field get contradictory treatment. The existing test `CommunityCodecTests.Parse_MasksValueTo16Bits` legitimizes the truncation and would need to be inverted together with the fix.

## Execution path and impact

Operator-supplied community strings reach the wire through this codec from config sources (`RouteSeedingService.SeedAsync`, seeding tag) and peer custom sources (stored via the API, resolved at send time by `ConfigCommunityResolver`). A typo like `"65444:99999"` becomes `65444:34463` with no warning — the peer's incoming filters then match nothing the operator configured, and nothing on either side says why.

Interplay with #266 item 3 (AddSource community validation): if that validator delegates to `Parse`/`TryParse`, it will **pass** `"65444:99999"` — validation does not catch this today; the codec must reject it.

## Fix

Symmetric check:

```csharp
if (value > 0xFFFF)
    throw new FormatException($"Invalid community '{community}': VALUE part must be 0-65535 (got {value}).");
```

## Test

`Parse("1:131071")` and `Parse("65000:65536")` → `FormatException`; `Parse("65000:65535")` still parses. Update `Parse_MasksValueTo16Bits` to expect the throw.

ruhex:
Closes #328

## Problem

`CommunityCodec.Parse` validated the ASN half of `"ASN:VALUE"` but silently masked the VALUE half (`value & 0xFFFF`): `"65000:70000"` became `65000:4464` with no error. Operator-supplied tags from config sources and peer custom sources went to the wire silently wrong, and nothing on either side said why. The existing fixture `Parse_MasksValueTo16Bits` legitimized the truncation.

## Fix

1. **Symmetric rejection** in `Parse`: `value > 0xFFFF` → `FormatException`, mirroring the ASN branch; the mask is gone (value is already ≤ 0xFFFF). The legitimizing fixture is inverted into the regression test for the throw (that inversion is the red proof).
2. **`uint.Parse` → `uint.TryParse`** (review finding): huge/negative/non-numeric parts used to surface as `OverflowException`, which no caller's `FormatException` filter catches — the ctor path of `ConfigCommunityResolver` could crash host build on `"65000:99999999999999999999"`. Every malformed input is now a `FormatException`.
3. **Seeding companion** (`RouteSeedingService`, the runtime half of #327): the bare `Parse` in the per-source loop meant the new throw (and the pre-existing ASN-half throw) would abort the whole seeding run — later sources unseeded, warm-up and the established-session push skipped. The parse now degrades that one source to **untagged** with a Warning (the `ConfigCommunityResolver` never-throw contract), and the summary log reflects the *applied* tag, not the configured one. Fail-loud `Validate()` for `PrefixSources` stays in #327.

Callers: `ConfigCommunityResolver` (both sites catch `FormatException` → Warning + fallback — now also correctly rejecting 65536+ values they previously mistagged); `RouteSeedingService` (guarded, see above). `HandleAddSource` stores the raw string — its API-side 400 validation is #266 item 3 and now gets the reject for free once it delegates to the codec.

## Verification

- `dotnet test` — **768/768**.
- New tests: `Parse_ValueAbove65535_Throws` (+ boundary `Parse_MaxValue65535_Accepted`), `Parse_NonNumericOrHugeParts_ThrowFormatException_NotOverflow`, and `BadSourceCommunity_DegradesToUntagged_DoesNotAbortSeeding` (two sources, first with `65444:99999` → both seeded, first untagged, second correctly tagged).
- Internal reviewer: two passes — first found the seeding regression the strict codec would introduce (fixed in this PR as companion 3) and the OverflowException asymmetry (fixed as companion 2); second pass confirmed everything closed except one MINOR (summary-log lie after degrade — fixed as well).

## Notes

- Until #327's `Validate()` half lands, a bad community in config is still accepted at startup and only degrades at seeding with a Warning — startup fail-loud is tracked there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid community values are now rejected with consistent format errors.
  - Values above the valid 16-bit range are no longer silently truncated.
  - Route seeding continues when a source has an invalid community; that source is seeded without tags while valid sources continue processing.
  - Community status and logging now accurately reflect the applied result.

### DIFF
--- a/BGPLite.Protocol/CommunityCodec.cs
+++ b/BGPLite.Protocol/CommunityCodec.cs
@@ -12,13 +12,17 @@ public static class CommunityCodec
         if (colon < 0)
             throw new FormatException($"Invalid community '{community}' (expected 'ASN:VALUE').");
 
-        var asn = uint.Parse(community[..colon]);
-        var value = uint.Parse(community[(colon + 1)..]);
+        // uint.Parse would surface huge/negative parts as OverflowException, which no caller's
+        // FormatException filter catches — TryParse keeps every malformed input a FormatException.
+        if (!uint.TryParse(community[..colon], out var asn) || !uint.TryParse(community[(colon + 1)..], out var value))
+            throw new FormatException($"Invalid community '{community}': ASN and VALUE must be non-negative integers.");
 
         if (asn > 0xFFFF)
             throw new FormatException($"Invalid community '{community}': ASN part must be 0-65535 (got {asn}).");
+        if (value > 0xFFFF)
+            throw new FormatException($"Invalid community '{community}': VALUE part must be 0-65535 (got {value}).");
 
-        return (asn << 16) | (value & 0xFFFF);
+        return (asn << 16) | value;
     }
 
     public static string Format(uint community) => $"{community >> 16}:{community & 0xFFFF}";
--- a/BGPLite.Tests/CommunityCodecTests.cs
+++ b/BGPLite.Tests/CommunityCodecTests.cs
@@ -24,10 +24,28 @@ public class CommunityCodecTests
     }
 
     [Fact]
-    public void Parse_MasksValueTo16Bits()
+    public void Parse_ValueAbove65535_Throws()
     {
-        // 131071 = 0x1FFFF → low 16 bits 0xFFFF
-        Assert.Equal((1u << 16) | 0xFFFFu, CommunityCodec.Parse("1:131071"));
+        // #328: the VALUE half used to be silently masked (131071 = 0x1FFFF → low 16 bits 0xFFFF),
+        // corrupting operator-supplied tags without a trace — symmetric with the ASN check now.
+        var ex = Assert.Throws<FormatException>(() => CommunityCodec.Parse("1:131071"));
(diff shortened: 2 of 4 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

Issue #230, opened 2026-08-03
architecture: BGPLite.Api references BGPLite.Server — HTTP layer depends on FSM/transport (layering violation)


ruhex:
## What is wrong

`BGPLite.Api` has a direct `ProjectReference` to `BGPLite.Server`, so the HTTP management layer depends on the BGP FSM/transport layer:

```
BGPLite.Api/BGPLite.Api.csproj:18   <ProjectReference Include="..\BGPLite.Server\BGPLite.Server.csproj" />
BGPLite.Api/ManagementApi.cs:12      using BGPLite.Server;
BGPLite.Api/ManagementApi.cs:33      private readonly BgpMetrics _metrics;            // Server infrastructure type
BGPLite.Api/ManagementApi.cs:49      ISessionManager? sessionManager = null            // implemented by BgpServer
BGPLite/Program.cs:176               builder.Services.AddSingleton<ISessionManager>(sp => bgpServer!);
```

`ISessionManager` (`BGPLite.Server/ISessionManager.cs`) is implemented by `BgpServer` — the TCP listener and FSM owner. `BgpMetrics` is a Server-internal metrics object. So `ManagementApi` (an HTTP/edge concern) reaches across into the transport/FSM layer, inverting the intended dependency direction.

This violates the layering rule in `AGENTS.md`:

> Keep the boundary clear: protocol correctness belongs in `BGPLite.Protocol`; session FSM and TCP handling belong in `BGPLite.Server`; routing and filtering belong in `BGPLite.Routing`; HTTP management and persistence belong in `BGPLite.Api`.

**Impact:**
- The Api project transitively pulls in all of Server's concerns (FSM, timers, accept loop, connection throttle). A change to the FSM can break Api compilation.
- It blocks testability: `ManagementApi` cannot be unit-tested with a mock session manager without dragging in the Server assembly.
- It is inconsistent with the rest of the layering, where lower layers (Routing/Configuration) do not reference Server.

## Fix

Introduce a narrow read+refresh contract in a neutral layer and have Api depend on **that** instead of `BGPLite.Server`:

[...]

Issue #89, opened 2026-07-03
[arch/critical] config: AppConfig/BgpConfig are plain records bound from YAML with N


ruhex:
From the architecture + .NET 10/C# best-practices review (2026-07-03, 12 aspect-agents, 83 findings).

## Problem (CRITICAL, high)
**`BGPLite/BGPLite.Configuration/BgpConfig.cs:12`** (di-lifecycle-config)

AppConfig/BgpConfig are plain records bound from YAML with NO validation and RFC-invalid defaults: RouterId defaults to "0.0.0.0" (RFC 4271 §6.8 forbids this — the peer-side validator at BgpSession.cs:1211 rejects an all-zero BGP Identifier, yet the local side advertises it), Asn defaults to 0, HoldTime/KeepAlive have no range checks, and ApiPort has no bounds check. GetRouterIdAddress() (BgpConfig.cs:38) calls IPAddress.Parse which only throws on a malformed string, never on 0.0.0.0. ConfigLoader.IgnoreUnmatchedProperties means a typo like `Holdtime:` or `RouteRid:` is silently dropped and the app runs on bad defaults. Result: the route server binds port 179, opens the management API, and starts advertising before any error surfaces — the first BGP OPEN from a real peer is what rejects our 0.0.0.0 RouterId (BadBgpIdentifier) / Asn=0. This is the longest possible fail distance from cause to symptom and the worst possible time to discover misconfiguration.

## Fix
Add an AppConfig.Validate() (or IValidateOptions<AppConfig>) that runs in ConfigLoader.Load/LoadFromText immediately after deserialization and throws before the host is built. Rules: Asn in [1, 4294967295]; RouterId parses to a non-zero IPv4 (reject 0.0.0.0, reject IPv6); HoldTime ∈ {0} ∪ [3,65535]; 1 ≤ KeepAlive ≤ HoldTime/3 (or compute KeepAlive from HoldTime when unset); ApiPort in (0,65535]; each PeerConfig.Address parses to non-zero IPv4. Consider switching IgnoreUnmatchedProperties off (or at least logging unknown keys) so silent typos surface. This is FIXPLAN P4 (4.1–4.5) and remains unimplemented.

PR #336, opened 2026-08-27
fix(config): validate PrefixSources and every config community at startup (#327)

### LINKED ISSUE #327: config: a malformed PrefixSources Community aborts the whole seeding loop and Validate() never checks PrefixSources

ruhex:
Closes #327

## Problem

`AppConfig.Validate()` never looked at `PrefixSources`. A bad community, an unknown `Kind`, a missing `Path`/`Url`/`Asn`, a non-positive `Timeout`, or an empty/duplicate name surfaced only at load time — where `LoadAllAsync` absorbs the failure into a Warning plus an empty prefix set. A config typo silently served zero prefixes until restart, and `DefaultPrefixSource` typos ("ruu" vs "ru") emptied the default set for every unconfigured peer the same way. This violates the fail-loud policy (#89/#8) and is the fail-loud half of #327; the runtime half (per-source degrade to untagged in seeding) landed in #335.

## Fix — `Validate()` now checks

- **Per-kind required fields** (file→`Path`, http→`Url`, asn→`Asn`) — mirroring the providers' own load-time checks;
- **`Kind` ∈ {file, http, asn}** (case-sensitive, exactly as the factory resolves it) and **absolute http(s) URLs** for http sources (a malformed URL previously died inside HttpClient post-startup → silent zero prefixes; SSRF depth stays at fetch time in `PrefixSourceUrlValidator`);
- **AS 0 rejected** for asn sources (RFC 7607);
- **`Timeout` null or positive** (the provider silently ignores ≤ 0);
- **Non-empty, unique names** (sources are addressed by name — subscriptions, per-source cache);
- **`DefaultPrefixSource` must name a configured source**;
- **Every community string in the file** — `PrefixSources[].Community`, `CustomPrefixCommunity`, `CustomAsnCommunity`, `RipeStat.AsnLists[].Community` — through one rule via `CommunityCodec` (with the inner `FormatException` preserved). For that single-sourcing, `BGPLite.Configuration` now references the `BGPLite.Protocol` leaf — an edge the #230 layering matrix allows (only Configuration→{Api,Server,Routing,Providers,Contracts} is forbidden); `LayeringTests` stay green, `docs/ARCHITECTURE.md` updated to match (plus a pre-existing Server-row package inaccuracy fixed).
- Explicit YAML null collections (`PrefixSources:`, `AsnLists:` with no value) keep meaning "none" instead of NRE-ing the validator — regression-tested through the real `ConfigLoader`.

Hot-reload semantics are unchanged in shape: `ConfigReloader` validates before applying, so a bad edit is rejected and the previous config keeps running — same as every other invalid field.

## Verification

- `dotnet test` — **787/787** (16 new cases incl. theories).
- Internal reviewer: two passes — first found the missing `DefaultPrefixSource` check, the URL-format gap, the stale `ARCHITECTURE.md` table, AS 0, community fragmentation across the file, and the missing inner exception; second pass confirmed all closed and flagged YAML-null collections + the Server row — both fixed (the null-scalar behavior is pinned by a `ConfigLoader`-level test).
- `appsettings.yml` / `appsettings.Example.yml` pass the new validation (checked source-by-source by the reviewer).

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive configuration validation for prefix sources, including source types, required values, timeouts, unique names, and default source references.
  - Added validation for community values across custom and RipeStat settings.
  - Invalid or unsupported configuration values now produce clear validation failures.

- **Documentation**
  - Updated architecture documentation to reflect configuration validation dependencies and project relationships.

### DIFF
--- a/BGPLite.Configuration/AppConfig.cs
+++ b/BGPLite.Configuration/AppConfig.cs
@@ -1,5 +1,6 @@
 using System.Net;
 using System.Net.Sockets;
+using BGPLite.Protocol;
 using YamlDotNet.Serialization;
 
 namespace BGPLite.Configuration;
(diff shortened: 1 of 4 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
The MINOR above (null list elements) is fixed in 27605f6 exactly as suggested — indexed rejection with a message, plus two regression tests. The re-review of that commit could not run: the free-OSS review quota stayed exhausted through 7 retry triggers over ~35 minutes, and the maintainer opted to merge per the #334 precedent. Compensating gates on 27605f6: green CI (build-test / format-check / analyze / CodeQL), 789/789 locally, and the internal adversarial review already covered null-element handling in its second pass (its YAML-null-collection finding was the sibling case). Any post-merge CodeRabbit findings will be handled as follow-ups.

PR #337, opened 2026-08-28
fix(api): cancel in-flight handlers on shutdown and bound the drain (#326)

### LINKED ISSUE #326: api: ManagementApi.StopAsync ignores its token and in-flight handlers cannot be cancelled — the shutdown drain can starve the BGP Cease teardown

## Problem

`StopAsync` (`ManagementApi.cs:181-198`) never touches its `cancellationToken` and waits for in-flight handlers with no bound:

```csharp
public async Task StopAsync(CancellationToken cancellationToken)
{
    _cts.Cancel();               // cancels only the accept loop, not handlers
    ...
    if (pending.Length > 0) { try { await Task.WhenAll(pending); } catch { } }   // no timeout, no ct
}
```

Handlers have no shutdown observability: `CollectPeerPrefixes` / `HandleGetAsnListsAsync` call `_prefixService.GetPrefixesForAsns(...)` / `GetRuPrefixesAsync()` **without a token** (`ManagementApi.cs:941, 953, 983-989`), while the RIPEstat per-attempt timeout defaults to `TimeoutSeconds` = 180 s with `RetryAttempts` backoff, and `/api/asn-lists` iterates every ASN of every list sequentially — the total drain time is unbounded. (`HandleAsync` does propagate OCE correctly, but the token it would need is never cancelled.)

## Execution path and impact

Hosted services stop in reverse registration order (`Program.cs:209-236`): PrefixAutoRefreshService → ConfigReloader → **ManagementApi** → **BgpServer** → RouteSeedingService. So `BgpServer.StopAsync` — the Cease NOTIFICATION + socket teardown — waits for the management API's drain. A single `GET /api/peers/{id}/prefixes` or `GET /api/asn-lists` against a cold RIPEstat cache can hold shutdown for minutes; the host `ShutdownTimeout` (30 s default) only cancels a token nobody here listens to. In Docker (default 10 s stop grace) the process is SIGKILLed: peers get TCP RST instead of Cease, contradicting the "exactly one NOTIFICATION before close" teardown guarantee.

(Server-side analog was #161, fixed for `BgpServer`; this is the same class on the API surface. Related shape: #262 for the sync store underneath.)

## Fix

1. Add a handler-scoped CTS cancelled at the top of `StopAsync`; flow it through `RouteAsync` → handlers → the provider calls (the `ct` signatures already exist on `IPrefixService`).
2. Bound the drain: `await Task.WhenAny(Task.WhenAll(pending), Task.Delay(TimeSpan.FromSeconds(10), cancellationToken));`

## Test

Handler blocked on a `TaskCompletionSource` + `StopAsync` called with an already-cancelled token → assert `StopAsync` completes in < 1 s (today it hangs until the handler finishes); cancelled provider calls surface as 503/closed response, not a faulted unobserved task.

ruhex:
Closes #326

## Problem

`StopAsync` ignored its `CancellationToken` and awaited in-flight handlers via an unbounded `Task.WhenAll`. Handlers made provider calls with no token — `CollectPeerPrefixes` (`/api/peers/{id}/prefixes`), `HandleGetAsnListsAsync` (`/api/asn-lists`, sequential RIPEstat per ASN), `HandleGetAsnPrefixes` (`?count=true`), and `LoadAllAsync` in asn-lists (cold http prefix sources, 5 s x 3 attempts each). Worst case: minutes. The host stops services in reverse registration order, so `BgpServer.StopAsync` (Cease + socket teardown) waited behind the drain; in Docker (10 s stop grace) the process was SIGKILLed — peers saw TCP RST instead of Cease.

## Fix

- `_shutdownCts`, cancelled at the very top of `StopAsync` (deliberately never disposed: handlers abandoned by the bounded drain may still hold its token — a disposed source's `Token` getter would throw ODE at them).
- Every provider-calling handler observes the token (`ct` signatures already existed on `IPrefixService`/`IPrefixSourceService`); OCE is rethrown per the #114 policy — shutdown is not a fetch failure — and `HandleWithInflightReleaseAsync` no longer logs the shutdown unwind as `Management API request handling faulted`.
- Bounded drain: `WhenAny(WhenAll(pending), Task.Delay(10 s, hostToken))` — 10 s after cancellation, short of the host's 30 s grace.

## Verification

- `dotnet test` — **770/770**.
- Regression test drives a REAL HttpListener on a probed free port (with a TOCTOU retry): `GET /api/asn-lists` parks inside a provider that honors the token (`Task.Delay(Timeout.Infinite, ct)`); `StopAsync` completes in well under 5 s. Verified **red** on the old behavior (temporarily reverted: `TimeoutException` — StopAsync hung behind the uncancelled provider). A `Task.Yield` barrier before the provider signals guarantees the handler is registered in the in-flight list before the test proceeds.
- Internal reviewer found two provider calls the first pass missed (asn-count, LoadAllAsync), the OCE-as-fault log noise, and three test robustness issues — all fixed in this commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved API shutdown behavior by canceling in-progress operations and allowing active requests to finish within a bounded time.
  - Prevented expected shutdown cancellations from being reported as fetch failures.
  - Ensured prefix-related requests stop promptly when the service is shutting down.

- **Tests**
  - Added coverage for graceful shutdown during active provider requests.
  - Added coverage for startup behavior when the selected port is already in use.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -50,6 +50,12 @@ public sealed class ManagementApi : IHostedService, IDisposable
     private HttpListener? _listener;
     private Task? _listenTask;
     private readonly CancellationTokenSource _cts = new();
+    // #326: cancelled at the top of StopAsync so in-flight handlers stop provider work (RIPEstat
+    // fetches) instead of pinning the drain — the host stops services in reverse registration
+    // order, and BgpServer.StopAsync (the Cease + socket teardown) waits behind this drain.
+    // Deliberately never disposed: handlers abandoned by the bounded drain may still hold its
+    // token, and a disposed source's Token getter would throw ODE at them.
+    private readonly CancellationTokenSource _shutdownCts = new();
 
     public ManagementApi(
         PeerStore store,
@@ -180,6 +186,13 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
     public async Task StopAsync(CancellationToken cancellationToken)
     {
+        // #326: cancel in-flight handlers FIRST — they observe this token on provider calls, so a
+        // cold-cache RIPEstat fetch (per-attempt timeout 180 s × retries, /api/asn-lists sequential
+        // per ASN) unwinds immediately instead of pinning the drain. The host stops services in
+        // reverse registration order, so BgpServer.StopAsync (the Cease teardown) waits behind this
+        // method; an unbounded drain got the process SIGKILLed in Docker (default 10 s stop grace)
+        // — peers saw a TCP RST instead of the promised Cease.
+        _shutdownCts.Cancel();
         _cts.Cancel();
         _listener?.Stop();
         if (_listenTask is not null)
@@ -187,13 +200,14 @@ public sealed class ManagementApi : IHostedService, IDisposable
             try { await _listenTask; } catch { }
         }
 
-        // Drain in-flight handlers before the host disposes this service: a handler reaching
-        // _inflightCap.Release() after Dispose would hit ObjectDisposedException (#248 review).
+        // Drain in-flight handlers, but bounded: 10 s after cancellation is enough for a response
+        // write or an orderly OCE unwind, and short of the host's 30 s shutdown grace.
         Task[] pending;
         lock (_inflightHandlers) pending = _inflightHandlers.ToArray();
         if (pending.Length > 0)
         {
-            try { await Task.WhenAll(pending); } catch { }
+            try { await Task.WhenAny(Task.WhenAll(pending), Task.Delay(TimeSpan.FromSeconds(10), cancellationToken)); }
+            catch { /* host grace elapsed — proceed */ }
         }
     }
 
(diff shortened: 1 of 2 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
Both MINORs are applied verbatim in b1a1ad3 (guard the rethrows on the shutdown token; cancel the token in Dispose) — CI green, 770/770. The incremental re-review of that one-commit delta hit the quota again; merging per the established precedent (#334/#336/#338): the unreviewed delta is the reviewer's own suggested changes.

PR #338, opened 2026-08-28
fix(protocol): reject unsupported OPEN optional parameters with 2/4 (#329)

### LINKED ISSUE #329: protocol: an unrecognized OPEN Optional Parameter is silently skipped — no Unsupported Optional Parameter NOTIFICATION (RFC 4271 §6.2 subcode 4)

## Problem

`BgpMessageReader.ParseOptParameters` (`BgpMessageReader.cs:141-163`) handles only `paramType == 2` (Capabilities) and silently advances past everything else:

```csharp
if (paramType == 2) // Capability
    ParseCapabilities(data[offset..][..paramLen], capabilities);
offset += paramLen;    // any other type: silently ignored
```

RFC 4271 §6.2: an optional parameter that is not recognized must be answered with NOTIFICATION Open Message Error / **Unsupported Optional Parameter (subcode 4)**. The subcode constant does not even exist — `BgpConstants.SubError` jumps 3 → 6 (`BgpConstants.cs:49-53`).

Practical case: RFC 9072 Extended Optional Parameters (type 255) or an AUTH_INFO parameter — the sending speaker believes the parameter was accepted and acts on that assumption.

Scope note: RFC 5492 §4.2 "ignore unrecognized capabilities" applies only to capability TLVs *inside* a type-2 parameter — that part is implemented correctly and is unaffected by this fix. `docs/RFC_COMPLIANCE.md` already flags this area as partial (snapshot of 2026-07-02).

## Fix

Add `SubError.UnsupportedOptionalParameter = 4` and throw for non-2 parameter types:

```csharp
else throw new BgpParseException(
    $"Unsupported OPEN optional parameter type {paramType}",
    BgpConstants.Error.OpenMessageError, BgpConstants.SubError.UnsupportedOptionalParameter);
```

(optionally with the rejected `[paramType, paramLen]` as NOTIFICATION data, mirroring the malformed-capability handling).

## Test

OPEN carrying an Optional Parameter of type 1 (well-formed TLV) → `BgpNotificationException` 2/4 on the wire; OPEN with only type 2 still negotiates unchanged.

ruhex:
Closes #329

## Problem

`ParseOptParameters` handled only `paramType == 2` (Capabilities) and silently skipped every other type — a peer sending RFC 9072 Extended Optional Parameters (type 255) or the deprecated Authentication parameter (type 1) got silent acceptance instead of the RFC 4271 §6.2-mandated NOTIFICATION Open Message Error / Unsupported Optional Parameter (subcode 4). The subcode constant did not exist at all (`SubError` jumped 3 → 6).

## Fix

A well-formed TLV of an unknown parameter type now throws `BgpParseException(OpenMessageError, UnsupportedOptionalParameter)`. The session's OPEN receive path routes it to a NOTIFICATION and teardown — the exact path `UnsupportedVersion` already takes. Scope guards: unrecognized *capabilities* inside type 2 stay ignored (RFC 5492 §4.2, untouched); truncated TLV framing still reports 2/0 (bounds checks run before the type dispatch); the production writer only ever emits type 2, so no legitimate sender is broken.

## Verification

- `dotnet test` — **791/791** (2 new: type-1 and type-255 well-formed TLVs → 2/4; built with the existing `BuildRawOpen` raw-frame helper so framing is exercised, not the object model).
- Red is mechanical: before the change a non-type-2 TLV was skipped and `ReadMessage` succeeded, so `Assert.Throws` failed.
- Internal reviewer: PRISTINE (session routing to NOTIFICATION verified against the UnsupportedVersion path; no test or writer emits type ≠ 2).

## Summary by CodeRabbit

* **Bug Fixes**
  * OPEN messages containing unsupported optional parameter types are now rejected with the appropriate BGP error and subcode.
  * Supported capability parameters continue to be processed normally.

* **Tests**
  * Added coverage for unsupported authentication and extended optional parameter types.

### DIFF
--- a/BGPLite.Protocol/BgpConstants.cs
+++ b/BGPLite.Protocol/BgpConstants.cs
@@ -49,6 +49,7 @@ public static class BgpConstants
         public const byte UnsupportedVersion = 1;
         public const byte BadPeerAs = 2;
         public const byte BadBgpIdentifier = 3;
+        public const byte UnsupportedOptionalParameter = 4;
         public const byte UnacceptableHoldTime = 6;
         public const byte UnsupportedCapability = 7;
 
--- a/BGPLite.Protocol/BgpMessageReader.cs
+++ b/BGPLite.Protocol/BgpMessageReader.cs
@@ -156,7 +156,19 @@ public static class BgpMessageReader
                     BgpConstants.Error.OpenMessageError, BgpConstants.SubError.Unspecific);
 
             if (paramType == 2) // Capability
+            {
                 ParseCapabilities(data[offset..][..paramLen], capabilities);
+            }
+            else
+            {
+                // RFC 4271 §6.2 (#329): an optional parameter type this speaker does not recognize
+                // must be answered with Unsupported Optional Parameter (2/4) — the sender otherwise
+                // believes the parameter was accepted (e.g. RFC 9072 Extended Optional Parameters,
+                // type 255). Unrecognized CAPABILITIES inside type 2 stay ignored per RFC 5492 §4.2.
+                throw new BgpParseException(
+                    $"Unsupported OPEN optional parameter type {paramType} (length {paramLen})",
+                    BgpConstants.Error.OpenMessageError, BgpConstants.SubError.UnsupportedOptionalParameter);
+            }
 
             offset += paramLen;
         }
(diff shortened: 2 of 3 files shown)


ruhex:
The single MINOR (realistic type-255 frame) is applied verbatim as the committable suggestion (7e655b3) — CI green. The incremental re-review of that one-commit delta hit the free-OSS quota again; merging per the established precedent (#334/#336) since the unreviewed delta is the reviewer's own suggested test change.

Issue #320, opened 2026-08-27
providers,server: user-source fetches have no body-read timeout — a slow-reading URL hangs the session's route push forever and wedges the refresh debounce


ruhex:
Verified against `main` @ `64ce3a6`.

## Problem

`HttpPrefixProvider.LoadAsync` only applies a timeout when the *configured* source carries one:

```csharp
// HttpPrefixProvider.cs:52-63
if (source.Timeout is int seconds && seconds > 0)
{
    timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(ct);
    timeoutCts.CancelAfter(TimeSpan.FromSeconds(seconds));
    linkedToken = timeoutCts.Token;
}
else
{
    linkedToken = ct;   // <- user sources always land here
}
```

Per-peer user URL sources construct their config **without** a `Timeout`:

```csharp
// PrefixService.cs:65-71
var source = new PrefixSourceConfig { Kind = "http", Name = name, Url = url, Community = community };
```

Two more facts complete the hole:

- the named `"http"` client sets `c.Timeout = Timeout.InfiniteTimeSpan` (Program.cs:102) — required for the Polly pipeline, but it means the client enforces nothing either;
- the Polly `AddTimeout(TimeSpan.FromSeconds(5))` (Program.cs:303) wraps `SendAsync` **up to the response headers** only. With `HttpCompletionOption.ResponseHeadersRead` (HttpPrefixProvider.cs:88), the body loop

```csharp
while ((read = await stream.ReadAsync(buffer, linkedToken)) > 0)   // HttpPrefixProvider.cs:114
```

runs **outside** the resilience pipeline, guarded solely by `linkedToken` — which for a user source is just the session's `_cts.Token`.

## Execution path and impact

Peer adds `http://evil.example/list` via `POST /api/peers/{id}/sources` → next `SendAllRoutesAsync` (establish or refresh) → `RouteAssembler.AddUserSourceRoutesAsync` → `GetUserSourcePrefixesAsync` → `UserSourceCache.GetOrLoadAsync` takes the per-URL gate → server answers headers instantly, then drips the body one byte at a time:

1. `ReadAsync` blocks **indefinitely** — the session stays Established (the peer keeps sending KEEPALIVEs; the hold timer never fires), so the token never cancels. A merely slow/big list with no configured timeout behaves identically — no malice required.
[...]

Issue #285, opened 2026-08-26
server: send-budget timeout leaves a truncated frame on the wire and RefreshCycleAsync swallows the IOException — session stays Established on a desynced stream


ruhex:
## Summary

When the per-send budget in `SocketBgpConnection.WriteAsync` (#252/#279) fires, the in-flight socket write is **not** rolled back — .NET delivers whatever bytes it already handed to the kernel and abandons the rest. BGPLite then:

1. returns the pooled buffer and releases `_sendLock` as if the send had failed cleanly, and
2. in `BgpSession.RefreshCycleAsync`, **swallows** the resulting `IOException` in its generic `catch (Exception ex)` and leaves the session `Established`.

The peer is left with a truncated BGP frame at the head of its parse position, and BGPLite keeps writing complete frames after it. Every subsequent byte is consumed by the peer as payload of the truncated frame — permanent stream desync on a session both sides still believe is up.

## Impact

- **Silent route corruption / route loss.** The peer's BGP parser desyncs; the announcements it decodes from that point on are garbage or it resets the session on a marker error. Neither side logs a route problem.
- **Explains the class of symptom in #282**: garbled NOTIFICATION bytes on the wire with no code path that constructs them.
- The transport's own XML doc already states the invariant that is being violated: *"Aborting an in-flight socket write leaves the connection unusable"*. Nothing in the session acts on it.

## Affected code

`BGPLite.Server/SocketBgpConnection.cs` — `WriteAsync`:

```csharp
using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
timeoutCts.CancelAfter(_sendTimeoutMs);
try { await _stream.WriteAsync(buffer, timeoutCts.Token); }
...
catch (OperationCanceledException)
{
    throw new IOException($"Send timed out after {_sendTimeoutMs} ms — ...");
}
```

`BGPLite.Server/BgpSession.cs` — `RefreshCycleAsync`:

```csharp
try
{
    _logger.LogInformation("Refreshing routes for {Peer}", _peer);
    await WithdrawAllAsync();
    await SendAllRoutesAsync();
}
catch (OperationCanceledException) { /* shutdown / caller cancel — best effort */ }
catch (Exception ex)                                   // <-- IOException lands here
{
    _logger.LogError(ex, "Failed to refresh routes for {Peer}", _peer);
}
```

[...]

PR #340, opened 2026-08-28
chore: cross-layer audit minors — guarded OCE rethrows, retired-limiter disposal, response close, RIPEstat failure logging, doc drift (#330)

### LINKED ISSUE #330: chore: cross-layer audit minors — RouteAssembler OCE swallowing, stale seed routes, dispose-time send-lock waiter, rate-limiter dispose, response close, silent ASN fetch failures, doc drift
Grouped LOW/MINOR findings from the 2026-08-27 full-repo audit (verified against `main` @ `26d474e`). Sub-items can be split out if any grows. Higher-severity findings from the same audit were filed separately (#322–#329); duplicates of existing issues (#258, #266, #292) were not re-filed.

## 1. RouteAssembler swallows OperationCanceledException in seven catch-alls
`RouteAssembler.cs:87-90, 123-126, 144-148, 168-171, 204-208, 230-233, 252-255` — every `catch (Exception)` there also catches OCE. A session teardown that cancels `_cts` mid-`BuildOutboundRoutesAsync` logs a burst of **ERROR** "Failed to fetch ..." entries for what is a normal shutdown, masking real fetch failures. The same file already does it right at `:305` (`catch (OperationCanceledException) { throw; }`, "#114 propagation"). Fix: add the OCE clause to each block. Test: cancelled fetch → OCE propagates to `RefreshCycleAsync`'s OCE handler, zero LogError calls.

## 2. Seed routes are never refreshed after a source changes
The only writers to the shared `RouteTable` are startup seeding (`RouteSeedingService.cs:72`, unowned) and per-peer installs (`BgpSession.cs:892`, owned). When `PrefixAutoRefreshService`/`onSourceChanged` detects a content change, peers get the new prefixes pushed, but the **unowned seed rows keep the startup snapshot forever** — `GET /api/routes` (`ManagementApi.cs:1062`) and the D13 degraded-mode assembler (`SharedTableRouteAssembler`) serve stale data until restart. Fix: on detected change, re-seed that source's unowned rows (remove-by-source + AddOrUpdate), or document `/api/routes` as a startup snapshot. Test: change a source → after the refresh cycle the route table contains the new prefix.

## 3. Dispose-time send-lock waiter is never woken
`SendKeepaliveAsync` is the one send path without a token: it enters `_sendLock.WaitAsync(CancellationToken.None)` (`BgpSession.cs:1157` via `:1270`/`:813`). If `Dispose` runs while another sender holds `_sendLock` (slow-peer send inside the #285 budget), the queued waiter is never released — `SemaphoreSlim.Dispose` does not wake waiters, and the holder's `Release` in its finally is suppressed by `ObjectDisposedException` (`:1192-1193`). Result: `HoldTimerLoopAsync` never completes → `RunAsync` never returns, and the session graph stays rooted by the semaphore queue. Reachable only via external Dispose while `RunAsync` is live (the `StopAsync` dispose loop at `BgpServer.cs:132-135`). Today's practical impact is a hung task surviving process shutdown; it becomes real the moment the host lifecycle is reused (tests, embedded use). Fix: wait on `_cts.Token` instead of `None` (Dispose cancels `_cts` first; map OCE → `return false` like the other paths).

## 4. ApplyConfig retires rate limiters without Dispose
`ManagementApi.cs:111-117` swaps limiters and comments "Let GC collect them", but `TokenBucketRateLimiter` with `AutoReplenishment = true` roots a replenishment `System.Threading.Timer` that GC never collects (verified against dotnet/runtime `TokenBucketRateLimiter.cs`). Each hot reload leaks a ticking timer plus the partitioned limiter graph (per-IP partitions are never evicted either). The non-dispose is deliberate for #137 (in-flight acquires on the retired limiter), so dispose after drain — in `StopAsync`/`Dispose` — and correct both comments (the XML doc at `:94` already claims disposal happens). Test: N repeated `ApplyConfig` → retiring limiters eventually disposed (flag/mock).

## 5. HttpListenerResponse not closed on handler fault paths
`ManagementApi.cs:283, 303` write 429/503 outside the handler try; `:313-326` the catch path calls `WriteResponse` again, which can throw on an aborted client; `HandleWithInflightReleaseAsync` has no finally `Close()`. A client abort during a large export leaves the response to the finalizer. Fix: `try { ctx.Response.Close(); } catch { }` in the handler finally (Close is idempotent after a successful write). Test: throwing OutputStream → `Response.Close()` invoked.

## 6. Silent ASN prefix loss on RIPEstat failure
`PrefixService.ResolveAsnAsync` ends in `catch { return []; }` with no logging (`PrefixService.cs:251-255`), and the cold-failure negative-cache branch (`:114-117`) is equally silent (the stale branch at `:107-111` does warn — asymmetric). A transient RIPEstat outage makes one ASN's prefixes quietly vanish from advertisement; the only trace is a smaller count in an INFO line. The operator cannot distinguish "ASN no longer has prefixes" from "fetch failed". Fix: `LogWarning` with the exception in both places. Test: failing provider → a Warning is recorded.

## 7. Doc drift (two comments contradict code/tests and RFC text)
- `AttributeHelper.ReadLargeCommunities` docstring (`AttributeHelper.cs:168-172`) says a zero-length payload is a valid empty set, while code and `LargeCommunityCodecTests.Read_ZeroLength_Throws` reject it.
- `BgpMessageWriter.WriteUpdate` comment (`BgpMessageWriter.cs:149-153`) claims "RFC 4271 §5: well-known attributes MUST be sent ordered by type code" — no such requirement exists; D15 itself states ordering is an implementation invariant, not an RFC rule. Reword as the D15 invariant.

ruhex:
Closes #330 (items 1, 2, 4, 5, 6, 7 — item 3 deferred, see below)

Re-created from #339, which GitHub auto-closed when its stacked base branch (fix/326) was deleted by the #337 squash-merge; the diff is identical, rebased onto the merged #337 (cherry-pick applied cleanly — Dispose now carries both the #337 shutdown-cancel and this PR's retired-limiter disposal).

## Items

1. **RouteAssembler OCE swallowing** — all 7 per-fetch catch-alls rethrow OCE gated on `ct.IsCancellationRequested`: teardown cancellation propagates (#114) with no ERROR burst, while a per-source **timeout** OCE (HttpPrefixProvider's linked CTS, live caller token) stays a logged fetch failure — an *unguarded* rethrow would tear the session down / withdraw-all on one slow source (internal review BLOCKER). The pre-existing unguarded clause in `AddUserSourceRoutesAsync` is deliberately untouched: guarding it breaks the recorded #114 contract (`OperationCanceled_Propagates`); resolving that tension belongs with #320 (noted on both issues).
2. **Seed-table staleness** — documented at `GET /api/routes`: peer rows are live, unowned seed rows are a startup snapshot (the issue allowed fix-or-document).
4. **Retired limiter disposal** — swapped-out limiters are parked and disposed in StopAsync (after the bounded drain) and Dispose; the old "let GC collect them" comment was wrong (AutoReplenishment roots a Timer GC never collects). No dedicated test: no injectable creation seam, a reflection test would be brittle — 10 structural lines covered by build + existing hot-reload tests.
5. **Response close on fault paths** — the handler wrapper's finally closes the response (idempotent), so 429/503/throwing-WriteResponse paths no longer wait for the finalizer.
6. **Silent ASN failure** — `ResolveAsnAsync` logs a Warning when an ASN's fetch fails; previously its prefixes vanished from the cycle with no trace. Tested via the existing PerAsnHandler harness.
7. **Doc drift** — ReadLargeCommunities docstring matches code+test (zero-length is malformed); the writer's ordering comment states the D15 invariant instead of a nonexistent RFC 4271 MUST.

## Verification

- `dotnet test` — **796/796** on the rebased branch.
- Internal reviewer: three passes — the first found the per-source-timeout BLOCKER (fixed via the guard + two-sided tests), the third **PRISTINE**.

## Deferred

Item 3 (dispose-time send-lock waiter) — concurrency-sensitive (#285/#302 territory); deferred with a note and suggested fix direction on the issue.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -26,6 +26,10 @@ public sealed class ManagementApi : IHostedService, IDisposable
     private IReadOnlyList<IPNetwork> _trustedProxyNetworks;
     private PartitionedRateLimiter<string>? _rateLimiter;
     private ConcurrencyLimiter? _concurrencyLimiter;
+    // #330: limiters swapped out by ApplyConfig, disposed in StopAsync/Dispose after in-flight
+    // requests have drained — a retired TokenBucketRateLimiter with AutoReplenishment roots a
+    // replenish Timer that GC never collects, so "let GC handle it" leaked one timer per reload.
+    private readonly List<IDisposable> _retiredLimiters = [];
     private IReadOnlyList<string>? _corsAllowedOrigins;
     private readonly BgpMetrics _metrics;
     // #263: required, not optional. Each of these was a silent feature switch: without
@@ -114,13 +118,18 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // Swap every reloadable field atomically. A request that has already captured the old
         // references into locals finishes against them; the next request reads the new ones.
         // _config is swapped last so CORS / client-IP and the limiters always move together.
-        Interlocked.Exchange(ref _rateLimiter, rateLimiter);
-        Interlocked.Exchange(ref _concurrencyLimiter, concurrencyLimiter);
+        var oldRateLimiter = Interlocked.Exchange(ref _rateLimiter, rateLimiter);
+        var oldConcurrencyLimiter = Interlocked.Exchange(ref _concurrencyLimiter, concurrencyLimiter);
         Interlocked.Exchange(ref _trustedProxyNetworks, trusted);
         Interlocked.Exchange(ref _corsAllowedOrigins, newConfig.CorsAllowedOrigins);
 
-        // Old limiters are NOT disposed: a concurrent HandleAsync may have captured the old reference
-        // and be mid-acquire. Let GC collect them (reload is rare, one retired object per reload).
+        // Old limiters cannot be disposed here — a concurrent HandleAsync may still be mid-acquire
+        // on them (#137). Park them for StopAsync/Dispose, which run after the in-flight drain.
+        lock (_retiredLimiters)
+        {
+            if (oldRateLimiter is not null) _retiredLimiters.Add(oldRateLimiter);
+            if (oldConcurrencyLimiter is not null) _retiredLimiters.Add(oldConcurrencyLimiter);
+        }
 
         _logger.LogInformation(
             "Soft config reloaded: trustedProxies={TrustedProxyCount}, corsOrigins={CorsOriginCount}, rateLimit={RateLimitEnabled}, concurrencyLimit={ConcurrencyEnabled}",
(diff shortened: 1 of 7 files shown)


ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
CodeRabbit could not review this PR: the free-OSS quota stayed exhausted through 8 retry triggers over ~40 minutes (the evening's quota was consumed by #336/#338/#337 reviews). Merging per the established precedent (#334/#336/#338/#337). Compensating gates: three internal adversarial review passes (the first caught a real BLOCKER — an unguarded OCE rethrow would tear sessions down on a single slow source — fixed and re-reviewed to PRISTINE), green CI on build-test/format-check/analyze/CodeQL, 796/796 locally. Any post-merge CodeRabbit findings will be handled as follow-up commits.

Issue #324, opened 2026-08-27
providers: config prefix sources without a Timeout have no body-read deadline — one slow-dripping source wedges seeding and auto-refresh (extends #320)


ruhex:
Verified against `main` @ `26d474e`.

## Problem

#320 tracks the **user-source** half of the missing body-read deadline. This issue covers the half its proposed fix does not close: **YAML-configured `PrefixSources` entries that omit `Timeout`**.

The mechanism is the same as #320 (see there for the full trace):

```csharp
if (source.Timeout is int seconds && seconds > 0) { ...CancelAfter... }
else { linkedToken = ct; }                       // HttpPrefixProvider.cs:60-63
...
while ((read = await stream.ReadAsync(buffer, linkedToken)) > 0)   // :114
```

- the named `"http"` client sets `Timeout.InfiniteTimeSpan` (`Program.cs:102`);
- the Polly `AddTimeout(5s)` pipeline wraps `SendAsync` **only up to the response headers** — with `HttpCompletionOption.ResponseHeadersRead` (`HttpPrefixProvider.cs:88`) the body loop runs outside the pipeline;
- `MaxResponseBytes` (10 MB) bounds size, not time: a server that answers headers instantly and then drips one byte per minute hangs the read indefinitely.

## Execution path and impact (the part beyond #320)

Config sources have no construction site that could arm a default (unlike user sources) — whatever the YAML says is all there is:

- `RouteSeedingService.SeedAsync` → `PrefixSourceService.LoadAllAsync` awaits sources **sequentially**: one dripping source stalls every later source, `WarmUpAsync`, and the final `RefreshAllEstablishedAsync` (`RouteSeedingService.cs:62-103) — i.e. startup seeding is wedged until process restart, with listeners already up.
- `PrefixAutoRefreshService.CheckSourcesAsync` iterates per-source gates `_locks[name]` sequentially: one hung `RefreshAsync` stops change-detection for **all** sources until restart.

Interplay with #267 item 2 (pipeline 5s caps a *configured* `source.Timeout`): today the two failure modes are complementary — no `Timeout` configured → unbounded body read; `Timeout` configured → silently capped at 5 s by the shared pipeline anyway. Any fix should settle both at once.

## Fix

[...]

Issue #319, opened 2026-08-27
providers: RipeStatProvider bypasses the canonical PrefixCidr parser — no host-bit masking, no /0 rejection, no length range check (#162/#236 left one path open)


ruhex:
Verified against `main` @ `64ce3a6`.

## Problem

`RipeStatProvider.GetPrefixesAsync` parses RIPEstat's CIDR strings by hand, bypassing the canonical `PrefixCidr` parser every other input path uses:

```csharp
// RipeStatProvider.cs:51-58
var cidr = element.GetString()!;
var slash = cidr.IndexOf('/');
var ip = IPAddress.Parse(cidr[..slash]);
var length = byte.Parse(cidr[(slash + 1)..]);
var prefix = BgpConstants.IPAddressToUint(ip);
result.Add((prefix, length));
```

For comparison, every other prefix entry point goes through `PrefixCidr.TryParse` (ManagementApi custom prefixes, `PrefixListParser` for file/HTTP/user-URL sources, `RouteAssembler` for stored custom prefixes), which masks host bits, rejects `/0`, and enforces length 1..32 (#236; #162 fixed the same defect for `PrefixListParser`).

What the RIPEstat path accepts as-is:

```
"10.0.0.1/8"    -> prefix=0x0A000001, length=8   (host bits NOT masked)
"0.0.0.0/0"     -> prefix=0, length=0            (default route admitted)
"1.2.3.4/33"    -> length=33                     (no range check)
```

## Why the inputs are not trusted

RIPEstat's `ris-prefixes` returns what RIS collectors saw **announced**. A malicious/buggy third party announcing non-canonical NLRI (`10.0.0.1/8`) or `0.0.0.0/0` from an origin AS an operator subscribes to puts that string into this parser. Inbound BGP NLRIs are masked by `PrefixCodec.Decode` (`addr &= 0xFFFFFFFF << (32 - length)`), so BGPLite already acknowledges the class; this is the one path where it does not.

## Impact

- **Route-table key corruption**: `(0x0A000001, 8)` and a canonical `(0x0A000000, 8)` for the same prefix coexist in the `RouteTable` (keyed by `(Prefix, Length)`), breaking dedup and the aggregation contract — `ExactUnionPrefixAggregator` masks on its own, so the same route can appear twice on the wire under two keys.
[...]

Issue #158, opened 2026-07-05
[security/providers] SSRF bypass via 6to4/teredo IPv6 + no port restriction


ruhex:
## Bug

Two gaps in `PrefixSourceUrlValidator` (the #144 SSRF defense):

**1. IPv6 normalization is incomplete** — `IsBlockedAddress` maps IPv4-mapped-to-IPv6 but does not cover IPv6 forms that **embed a private IPv4**:

```csharp
// BGPLite.Providers/PrefixSourceUrlValidator.cs:39-45
internal static bool IsBlockedAddress(IPAddress address)
{
    var normalized = address.IsIPv4MappedToIPv6 ? address.MapToIPv4() : address;
    foreach (var range in BlockedRanges)
        if (range.Contains(normalized)) return true;
    return false;
}
```

Not blocked:
- `2002::/16` (6to4) — `2002:ac10:0001::` encodes `172.16.0.1` (RFC 1918 private).
- `2001::/32` (Teredo) — can embed private IPv4 in the last 32 bits.
- IPv4-compatible IPv6 (`::a.b.c.d`, deprecated but parseable).

An attacker controlling a DNS record can return such a 6to4 address for a peer-supplied URL and reach the connect path to an internal host.

**2. No port restriction** — `ValidateUrlAsync` checks scheme is http/https but does not restrict the port. A peer can supply `http://attacker:8080/...` or `http://internal-host:9000/...`. (The ConnectCallback validates the IP, but the port is attacker-controlled — internal services on non-standard ports remain reachable.) `ValidateUrlAsync` is also **not wired into the fetch path** — defense is solely at connect time.

## Impact

- SSRF bypass to internal IPv4-encoded IPv6 addresses (cloud-internal services, metadata endpoints via 6to4 encoding of 169.254.x.x).
- Internal port scanning / service access via unrestricted ports.
- Compounds #90 (no auth) — anyone can register a malicious source URL.

## Fix

- Add `2002::/16` (6to4), `2001::/32` (Teredo), `::/96` (IPv4-compatible) to `BlockedRanges`.
- For any IPv6 whose embedded IPv4 (last 32 bits for compatible/mapped forms) is private, reject.
- Restrict ports to `{80, 443}` (configurable allowlist) in `ValidateUrlAsync`.
- **Wire `ValidateUrlAsync` into `HttpPrefixProvider.LoadAsync`** so scheme/port validation runs on the live path (defense in depth alongside the ConnectCallback).

## Scope

[...]

PR #373, opened 2026-08-31
fix: providers/config audit minors — reload re-arm, redirect rejection, NAT64, RIPEstat bounds, async file reads, empty-YAML error, refresh teardown (#321)

### LINKED ISSUE #321: chore: providers/config audit minors — reload lost-update, redirect header re-send, NAT64 gap, RIPEstat parse robustness, sync file I/O, empty-YAML NRE, refresh dispose race
Grouped LOW findings from the 2026-08-27 providers/config audit (each verified against `main` @ `64ce3a6`; split out if any grows). Companion to #319/#320 (MEDIUM) and #261 (UserSourceCache, already tracked).

## 1. ConfigReloader: a save during an in-flight reload is lost
`ConfigReloader.cs:93-109` — the debounce timer is one-shot; when it fires while `_reloading == 1` the CAS fails and `TriggerReload` just returns, **without re-arming**. Sequence: save #1 → reload starts (slow) → save #2 → debounce fires mid-reload → CAS fail → timer dead → save #2's changes are never applied until the next save. Fix: in the CAS-failure branch `_debounceTimer.Change(DebounceMs, Timeout.Infinite)` (or a pending flag checked in the `finally`).

## 2. Redirect comments are wrong and non-Authorization headers leak on redirect
`Program.cs:107-108`, `HttpPrefixProvider.cs:15-17`, `PrefixSourceUrlValidator.cs:68-69` all claim "SocketsHttpHandler does not follow redirects". Verified against MS docs: `SocketsHttpHandler.AllowAutoRedirect` defaults to **true** — redirects ARE followed. The SSRF check still holds (each hop re-enters `ConnectCallback`, so the target IP/port are re-validated — good), but per-source request headers other than `Authorization` are **re-sent to the redirect target** (docs: only Authorization is cleared), so an operator header like `X-API-Key` set on a source whose host has an open redirect goes to the attacker's origin. Fix: set `AllowAutoRedirect = false` explicitly and reject 3xx (or strip headers on cross-origin redirect).

## 3. NAT64 well-known prefix missing from the SSRF blocklist
`PrefixSourceUrlValidator.cs:16-40` blocks 6to4/Teredo/mapped/compatible but not `64:ff9b::/96` (or `64:ff8:1::/96`). In a NAT64/DNS64 environment an attacker-controlled DNS name can resolve into the NAT64 prefix and be translated to an internal IPv4 — same class as the 2002::/16 bypass #158 closed. Also `ff00::/8` (benign, connect fails). Fix: add both prefixes to `IsBlocked`.

## 4. RIPEstat response parsing is fragile and unbounded
`RipeStatProvider.cs:40-58` — `ReadAsStringAsync(ct)` has no size cap (HttpPrefixProvider caps at 10 MB, #144); `element.GetString()!` NREs on a non-string; `cidr[..slash]` with `slash == -1` throws ArgumentOutOfRangeException; `GetProperty(...)` KeyNotFound on an unexpected 200-body. All surface as "fetch failed" (stale/negative cache) — wrong classification, not a crash. Fix: length-capped read + TryGetProperty chain + per-entry try/skip.

## 5. FilePrefixProvider: synchronous blocking I/O under the per-source gate
`FilePrefixProvider.cs:30-41` — `File.Exists` / `File.GetLastWriteTimeUtc` / `File.ReadAllText` run synchronously inside async `LoadAsync`, holding a threadpool thread under the per-source semaphore; `ct` is never used. Also `fileMtime == lastModified` equality can serve stale "304" on filesystems with coarse mtime granularity. Fix: `ReadAsStringAsync`-style async reads + `ct` + length-or-content check.

## 6. ConfigLoader: empty/whitespace YAML crashes as NRE instead of a clear error
`ConfigLoader.cs:13-17` — YamlDotNet 18.1.0 `Deserialize<AppConfig>("" | "   " | "---")` returns `null` (verified empirically); `Program.cs` then NREs at `config.Validate()` with an opaque stack instead of "configuration file is empty". Same in `ConfigReloader.Reload` (logged, previous config survives). Fix: null-check with a descriptive message.

## 7. PrefixAutoRefreshService: StopAsync disposes the timer a waiting loop still uses
`PrefixAutoRefreshService.cs:96-97` disposes `_timer` while the loop may be parked in `await _timer.WaitForNextTickAsync(ct)` (line 108, outside try/catch) → ObjectDisposedException fault, caught upstream as generic "loop faulted" log noise on every affected shutdown. Fix: wait the loop task (like BgpServer/ManagementApi do) before disposing, or wrap the wait in try/catch(ODE).

## 8. ManagementApi.ApplyConfig comments describe a `_config` swap that does not happen
`ManagementApi.cs:110` ("_config is swapped last") and line 94 vs 116-117 (limiters "disposed" vs "NOT disposed") contradict the code; hot-reload works because readers use the swapped derived fields. Doc-only, but it documents the wrong contract for the next editor.

## Acceptance
Per sub-item as described; items 1-3 have straightforward regression tests (reload re-arm; redirect header strip; NAT64 resolution rejection).

ruhex:
Closes #321

## Items (all eight; item 4's parse-robustness half landed with the #319 fix — this adds the size/time bounds)

1. **Reload lost-update** — the one-shot debounce timer re-arms on CAS failure: a save during an in-flight reload is applied once the reload finishes instead of being silently dropped. A `_stopped` latch (set before the timer is disarmed in StopAsync) keeps a queued callback from starting a reload during shutdown.
2. **Redirects** — `AllowAutoRedirect = false` on the named http handler. Verified against MS docs: the default is **true**, and on redirect the handler re-sends every per-source header except `Authorization` — an `X-API-Key` on a source whose host has an open redirect leaked to the attacker's origin. The SSRF gate itself held per hop (each connection re-enters `ConnectCallback`); the leak was the residual risk. `HttpPrefixProvider` now rejects 3xx (300–399, boundary-tested) with an operator-fixable "configure the final URL" error, and the three comments claiming "SocketsHttpHandler does not follow redirects" state reality.
3. **NAT64** — `64:ff9b::/96` and `64:ff8:1::/96` join the SSRF blocklist (a DNS64-resolved name can embed any internal IPv4; same class as the #158 6to4 bypass). Theory-extended.
4. **RIPEstat bounds** — the body is capped at 10 MB (fast ContentLength check + a hard streaming cap that catches a header-lying origin mid-stream — both tested) and time-bounded by a body deadline mirroring #324 (the resilience pipeline clips at headers under `ResponseHeadersRead`), clamped to a 10 s minimum like the pipeline so a misconfigured `TimeoutSeconds: 0` degrades to the minimum instead of cancelling every read.
5. **File reads** — `ReadAllTextAsync(ct)`; the sync read held a threadpool thread under the per-source gate.
6. **Empty YAML** — "configuration is empty" instead of an NRE; theory over `""` / `"   "` / `"---"`.
7. **Refresh teardown** — StopAsync awaits the loop (host-token-bounded, then a token-less 5 s join closing the host-grace race) before disposing the timer; the parked `WaitForNextTickAsync` no longer faults with an unobserved ODE.
8. **ApplyConfig comment** — no longer describes a `_config` swap that never happens (points at #266 item 6 for the restart-required fields).

## Verification

- `dotnet test` — **855/855** (9 new cases: redirect 302/300, NAT64 ×2, empty-YAML ×3, RIPEstat header-cap and streaming-cap).
- Internal reviewer: two passes — the first found five MINORs (3xx boundary at 300, the missing RIPEstat body deadline, the host-grace dispose race, an untested streaming path, the shutdown re-arm race); all fixed; the second confirmed everything closed except the `TimeoutSeconds: 0` clamp, applied after.

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty configuration files now produce a clear error.
  - Configuration reloads are more reliable during rapid edits and shutdown.
  - Application shutdown no longer generates misleading reload errors.
  - HTTP redirects from prefix sources are rejected clearly and are not followed automatically.

- **Security**
  - Expanded protection against SSRF attacks involving NAT64 addresses.
  - RIPEstat responses are limited to 10 MB to prevent excessive memory use.

- **Performance**
  - File-based configuration sources now load asynchronously with cancellation support.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -107,7 +107,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
     /// <summary>
     /// Hot-reloads the SOFT (non-session-disrupting) part of the configuration (#136): the
-    /// trusted-proxy CIDR list (client-IP resolution), the CORS origin allowlist (via <c>_config</c>),
+    /// trusted-proxy CIDR list (client-IP resolution), the CORS origin allowlist (<c>_corsAllowedOrigins</c>),
     /// and the API rate / concurrency limiters. Each derived field is rebuilt from
     /// <paramref name="newConfig"/> and swapped atomically with <see cref="Interlocked.Exchange"/> so
     /// in-flight requests keep observing the previous state while subsequent requests pick up the new
@@ -160,8 +160,8 @@ public sealed class ManagementApi : IHostedService, IDisposable
         ResolveClientIp(remote, xForwardedFor, xRealIp, Volatile.Read(ref _trustedProxyNetworks));
 
     /// <summary>
-    /// Resolves the CORS origin against the CURRENT live <c>_config</c> (#136), for tests that need
-    /// to observe the effect of reloading <c>CorsAllowedOrigins</c> without an HttpListener. Mirrors
+    /// Resolves the CORS origin against the CURRENT live <c>_corsAllowedOrigins</c> (#136), for tests
+    /// that need to observe the effect of reloading <c>CorsAllowedOrigins</c> without an HttpListener. Mirrors
     /// <see cref="AddCorsHeaders"/>'s resolution.
     /// </summary>
     internal string? ResolveCorsOriginLive(string? requestOrigin) =>
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -127,7 +127,9 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         // Swap every reloadable field atomically. A request that has already captured the old
         // references into locals finishes against them; the next request reads the new ones.
-        // _config is swapped last so CORS / client-IP and the limiters always move together.
+        // NOTE (#321 item 8): _config itself is NOT swapped — request-path code that must observe
+        // reloads reads the derived fields above; the fields still reading _config (e.g.
+        // MaxRequestBodyBytes, RipeStat lists) are restart-required, tracked as #266 item 6.
         var oldRateLimiter = Interlocked.Exchange(ref _rateLimiter, rateLimiter);
         var oldConcurrencyLimiter = Interlocked.Exchange(ref _concurrencyLimiter, concurrencyLimiter);
         Interlocked.Exchange(ref _trustedProxyNetworks, trusted);
--- a/BGPLite.Configuration/ConfigLoader.cs
+++ b/BGPLite.Configuration/ConfigLoader.cs
@@ -13,11 +13,15 @@ public static class ConfigLoader
     public static AppConfig Load(string path)
     {
         var yaml = File.ReadAllText(path);
-        return Deserializer.Deserialize<AppConfig>(yaml);
+        return LoadFromText(yaml);
     }
 
     public static AppConfig LoadFromText(string yaml) =>
-        Deserializer.Deserialize<AppConfig>(yaml);
+        // YamlDotNet returns null for an empty/whitespace document ("", "   ", "---") — fail loud
+        // with a clear message instead of an NRE at the first config use (#321 item 6). The
+        // hot-reload path logs it and keeps the previous config, as with any other bad edit.
+        Deserializer.Deserialize<AppConfig>(yaml)
+            ?? throw new InvalidOperationException("The configuration is empty — nothing to load.");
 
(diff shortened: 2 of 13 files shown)


ruhex:
Both MINORs applied verbatim in 9092d53 (stop/reload gate + StopAsync drain; doc-comment consistency) — CI green, 855/855. The incremental re-review hit the quota; merging per the established precedent: the unreviewed delta is the reviewer's own suggested changes.

Issue #390, opened 2026-09-01
config: validation leftovers from #8 — peer 0.0.0.0/RemoteAsn, RipeStat/AutoRefresh ints, ConfigLoader.Save fate


ruhex:
Split from #8 (config-validation audit, ~75% done). Three residual items:

## 1. Peer address/ASN validation gaps (AppConfig.Validate, Peers loop)
- `Peer.Address == "0.0.0.0"` parses as IPv4 and **passes** validation; per the #8 ask it should be rejected (RFC 4271 BGP Identifier analogy — a peer at 0.0.0.0 is never meaningful). Related: `PeerConfig.Address` still *defaults* to `"0.0.0.0"` (PeerConfig.cs:9) — an omitted Address silently becomes the invalid value; consider an empty default so omission trips validation.
- `PeerConfig.RemoteAsn` (`uint?`) is never checked — a peer row without an ASN is accepted and only fails later at session time. Decide: require non-null for configured peers (auto-registered peers are a separate path and unaffected).
- Tests: extend `Validate_RejectsBadPeerAddress` with the 0.0.0.0 case; add a RemoteAsn-required case.

## 2. RipeStat / AutoRefresh integer fields taken verbatim
- `RipeStatConfig.TimeoutSeconds` / `RetryAttempts` / `RetryDelaySeconds` (RipeStatConfig.cs:20/26/31) accept any int including 0/negative (0/negative RetryAttempts semantics are undefined; note the code path that reads them uses `Math.Max(10, …)` clamps for TimeoutSeconds — validation belongs at the boundary, not in clamps).
- `AutoRefreshConfig.IntervalSeconds` / `NoEtagIntervalSeconds` / `MaxJitterMs` (AutoRefreshConfig.cs:19-33) — same, no validation.
- Validate with clear messages + tests, mirroring the MaxRequestBodyBytes pattern (AppConfig.cs:112-115).

## 3. `ConfigLoader.Save` — delete or make atomic
- `ConfigLoader.Save(AppConfig)` (returns a YAML string) has **zero callers**: peers persist in SQLite and the YAML file is a read-only input (ConfigReloader only ever calls `Load`). The #8 item demanded an atomic `Save(path)` (temp + File.Replace) — that feature is obsolete unless an operator-facing "export config" is wanted.
- Recommendation: delete `Save` and its dead test expectations rather than implement; if an export feature is wanted, file it separately.

Refs: #8 (closed with per-item verdicts).

Issue #267, opened 2026-08-26
chore: providers/data audit minors — RIPEstat double buffering, pipeline timeout vs source.Timeout, EvictIfAtCapacity semaphore race, dead Includes, duplicate ASN caches


ruhex:
A grouping issue for LOW/MINOR providers/data findings from the 2026-08-26 audit.

## 1. RIPEstat response double buffering
`RipeStatProvider.cs:37-41` — `GetAsync` without `ResponseHeadersRead` + `ReadAsStringAsync` + `JsonDocument.Parse` ≈ 3× response size peak per fetch (UTF-16 string + document). Fix: `ResponseHeadersRead` + `ReadAsStreamAsync` + `ParseAsync`.

## 2. Pipeline timeout silently overrides per-source Timeout
`Program.cs:296` (`.AddTimeout(5s)` on the shared pipeline) vs `HttpPrefixProvider.cs:53-63` (linked token `CancelAfter(source.Timeout)`): an operator's `source.Timeout = 30` never gets more than 5 s. Fix: validate `Timeout <= pipeline budget` at config, or raise the pipeline timeout and rely on the linked token.

## 3. EvictIfAtCapacity can remove a semaphore someone holds
`PrefixService.cs:181-188` removes `_locks` entries without their gate: the holder keeps fetching while a new caller creates a second semaphore → duplicate concurrent fetch of one ASN (breaks the #164 invariant; cost = duplicate HTTP, not corruption). Also O(n²) `toEvict.Contains` (`:171`). Fix: evict locks only together with their cache entry and only when not held; `Contains` → HashSet.

## 4. Dead Includes
`PeerStore.cs:142-146` (`GetPeersByIp`: Include + non-entity Select — Include is ignored) and `:253-257, 264-268, 275-279` (GetCommunities*: Include + SelectMany — redundant). Fix: remove.

## 5. Duplicate ASN caches between mechanisms
The same ASN configured in both `RipeStat.AsnLists` (`PrefixService._cache`) and a `PrefixSources` entry of kind `asn` (`PrefixSourceService._cache`) is fetched and cached twice with independent TTLs (`AsnPrefixProvider.cs:8-10` documents this as design). For heavy overlap this doubles RIPEstat traffic. Fix (optional): share one cache keyed by ASN, or warn on config overlap.

## 6. Bonus roundtrip after create
[...]

Issue #229, opened 2026-08-03
providers: GetRuPrefixesAsync races on non-volatile _ruProjected/_ruCachedAt fields


ruhex:
## What is wrong

`PrefixService.GetRuPrefixesAsync` uses two plain (non-`volatile`, unsynchronized) fields shared across concurrent sessions:

```csharp
// BGPLite.Providers/PrefixService.cs:248-260
private List<(uint Prefix, byte Length, uint Asn)>? _ruProjected;
private DateTime _ruCachedAt;

public async Task<List<(uint Prefix, byte Length, uint Asn)>> GetRuPrefixesAsync(CancellationToken ct = default)
{
    if (_ruProjected is not null && _timeProvider.GetUtcNow().UtcDateTime - _ruCachedAt < _cacheTtl)
        return _ruProjected;
    var prefixes = await _prefixSources.GetDefaultAsync(ct);
    _ruProjected = prefixes.Select(p => (p.Prefix, p.Length, 0u)).ToList();   // unsynchronized write
    _ruCachedAt = _timeProvider.GetUtcNow().UtcDateTime;                       // unsynchronized write
    return _ruProjected;
}
```

`GetRuPrefixesAsync` is called concurrently from multiple BGP sessions (every peer whose routing policy resolves to the RU/default set hits this method), and `await` continuations can land on different thread-pool threads.

**Impact:**
- **Torn reads.** A reader can observe the newly-written `_ruProjected` reference together with the **old** `_ruCachedAt` value (the two writes are not ordered atomically). Without `volatile`/`Interlocked`, the JIT/CPU may reorder them. Result: the cache appears fresh-er or staler than it really is — a consistency violation, not data corruption, but it defeats the cache invariant.
- **Thundering herd.** Unlike the rest of `PrefixService` (`GetPrefixesAsync` uses a per-ASN `SemaphoreSlim` gate, `PrefixSourceService.LoadCachedAsync` uses a per-name gate), `GetRuPrefixesAsync` has **no gate**. When the TTL elapses, N concurrent callers all miss the cache and all invoke `_prefixSources.GetDefaultAsync(ct)` simultaneously — N duplicate loads of the same ~11k-prefix list. This is inconsistent with the resilience pattern enforced everywhere else in the file (see #164, #163).
- Inconsistency with the codebase policy: this is the only shared mutable cache state in `PrefixService` that is not protected the same way.

## Fix

Two options, consistent with the rest of the class:

[...]

PR #450, opened 2026-09-03
chore: integrate dev — external audit fixes #416–#431 (16 issues)

### LINKED ISSUE #416: providers: _ruGate deadlock — onSourceChanged push re-enters GetRuPrefixesAsync while the gate is held
## Found by
External deep audit of the codebase (2026-09-03) — concurrency review of the prefix-provisioning layer. Every link in the chain verified against source; not reproduced live.

## Summary
`PrefixService.GetRuPrefixesAsync` holds `_ruGate` across the whole default-source fetch. `PrefixSourceService.LoadCachedAsync` fires the `onSourceChanged` convergence callback INLINE after releasing only its own per-source gate, and that callback awaits `RefreshAllEstablishedAsync` → every established session's `SendAllRoutesAsync` → `BuildOutboundRoutesAsync` → `GetRuPrefixesAsync` — which blocks on the very gate the outer frame still holds. Any second RU-consuming established session turns this into a permanent circular wait.

## Deadlock chain
1. Session A refresh/build → `PrefixService.GetRuPrefixesAsync` acquires `_ruGate` (`BGPLite.Providers/PrefixService.cs:195`) → `await _prefixSources.GetDefaultAsync(ct)` (`PrefixService.cs:206`) — gate HELD across the await.
2. `PrefixSourceService.GetDefaultAsync` → `LoadCachedAsync(source, ct)` with `triggerCallback: true` (default). A first-ever load always counts as changed (`previousList is null`, `BGPLite.Providers/PrefixSourceService.cs:249`); after releasing the per-source gate the callback is awaited inline (`PrefixSourceService.cs:263-267`).
3. The callback awaits `ISessionManager.RefreshAllEstablishedAsync()` inline (`BGPLite/Program.cs:133-138`).
4. `BgpServer.RefreshAllEstablishedAsync` runs every established session's `RefreshRoutesAsync()` under `Task.WhenAll` (`BGPLite.Server/BgpServer.cs:517-538`). Session A returns instantly via the `_refreshRunning` debounce (`BgpSession.cs:134-138`), but session B enters `SendAllRoutesAsync` → `BuildOutboundRoutesAsync` → `GetRuPrefixesAsync` → `await _ruGate.WaitAsync(ct)` — blocked: A still holds the gate.
5. `WhenAll` waits B; B waits the gate; A waits `WhenAll`. B's token is its session token (healthy; KEEPALIVEs keep feeding the hold timer), so nothing cancels it. Permanent.

## Trigger
- `DefaultPrefixSource` configured, AND
- ≥ 2 established sessions whose outbound build consumes RU (unconfigured peers, 0-prefix RU fallback, country lists, auto-register), AND
- a cache-miss load that detects a change (both TTLs are 1h — TTL expiry + real content change; or failed seeding at startup with the source coming up later while peers are already established).

A single-peer deployment cannot reproduce it (the debounce saves the gate holder), which is also why tests never caught it: `PrefixCacheRaceTests` drives `PrefixService` with a FAKE `IPrefixSourceService`, so the cross-component re-entrancy is untested.

## Impact
Every later `GetRuPrefixesAsync` caller blocks forever: new unconfigured peers never complete their initial dump (handshake stalls after Established; the peer eventually gives up), all RU refreshes stall, management-API `RefreshPeer` requests hang. Sessions stay Established and logs stay quiet. Recovery = process restart.

## Fix shape
Do not run the convergence push on the gate-holder's stack while the RU gate is held. Cleanest: `IPrefixSourceService.GetDefaultAsync` returns `(prefixes, changed)` and `PrefixService` fires the push AFTER `_ruGate.Release()`. Alternatively detach the callback (background queue) — but then its failures must stay observable, not silently swallowed.

## Regression test
Real `PrefixService` + real `PrefixSourceService` over a fake provider whose second load returns a changed list; two scripted sessions whose `BuildOutboundRoutesAsync` calls `GetRuPrefixesAsync`; assert `RefreshAllEstablishedAsync` completes within a timeout (hangs today).

## Severity
CRITICAL — permanent, silent, process-wide stall of the primary route-provisioning path in the default multi-peer deployment profile.

## Related
- #214 (onSourceChanged convergence — introduced the inline callback)
- #229 (RU gate — introduced the single-gate serialization)

### LINKED ISSUE #417: providers: failed cold load of the default source is cached as an empty RU set for the full TTL
## Found by
External deep audit (2026-09-03) — prefix-provisioning resilience review.

## Summary
A single transient failure of the default prefix source during a cold RU load is cached as a POSITIVE empty result for the full RU cache TTL (1h default) — every unconfigured/unknown peer receives ZERO routes for up to an hour.

## Root cause
`PrefixSourceService.GetDefaultAsync` swallows every non-caller-cancel exception and returns `[]` (`BGPLite.Providers/PrefixSourceService.cs:95-99`). `PrefixService.GetRuPrefixesAsync` therefore never sees the failure: its stale-on-failure branch (`PrefixService.cs:210-224`) is dead code in production wiring — the empty list is projected and stored as a normal entry via `Interlocked.Exchange` (`PrefixService.cs:228`) for `_ruCacheTtl` (1h, `PrefixService.cs:36`).

The 30s negative cache inside `PrefixSourceService` does not help: the RU layer sits ON TOP of it and caches the empty projection positively.

`AutoRefresh` is disabled by default, so nothing repairs the cache early.

## Trigger
First (or post-TTL) load of the default source fails once — breaker open, 5xx across retries, network blip — while an unconfigured peer connects.

Note: `GetRuPrefixesAsync_FetchFailure_AfterExpiredCache_ServesStaleCopy` (`BGPLite.Tests/PrefixCacheRaceTests.cs:72`) validates the stale-on-failure behavior against a FAKE source service that throws — with the real `PrefixSourceService` that behavior is unreachable.

## Impact
Route loss for unconfigured peers (and peers on the 0-prefix RU fallback) for up to 1h, with every component looking healthy.

## Fix shape
Let `GetDefaultAsync` rethrow after logging (it already distinguishes caller-cancel at `:94`), so `PrefixService`'s stale-on-failure works as designed. If "source legitimately empty" must stay distinguishable from "fetch failed", propagate `(prefixes, failed)` instead of collapsing both into `[]`.

## Regression test
Real `PrefixService` + real `PrefixSourceService` with a provider that fails once then succeeds: assert the RU set is NOT cached as empty (throws to the caller / serves stale), and a later successful load repopulates within the same TTL window.

## Severity
HIGH — silent route loss on the RU data path.

## Related
- #163 (the stale-on-failure parity this dead code was written for)

### LINKED ISSUE #418: api,server: deleting/PATCHing one peer disarms the per-IP TCP-MD5 key for sibling peers
## Found by
External deep audit (2026-09-03) — management API review.

## Summary
The TCP-MD5 key table is keyed per source IP (`BGPLite.Server/BgpServer.cs:36`), but peer deletion and PATCH re-arm it from a SINGLE row. Deleting one of several peers sharing an IP removes the key for ALL of them — RFC 2385 enforcement is silently downgraded until a process restart re-runs the startup bootstrap.

## Root cause
- DELETE: `BGPLite.Api/ManagementApi.cs:1106` calls `_sessionManager.SetPeerMd5Key(peer.Ip, null)` unconditionally → `BgpServer.SetPeerMd5Key` removes the per-IP entry (`BgpServer.cs:182-187`).
- PATCH: `ManagementApi.cs:1071-1077` re-arms the whole-IP key from the just-edited row — setting/clearing one sibling's password overwrites (or clears) the sibling's key (last-writer-wins).
- The startup path itself documents that several peer rows can share one IP with different keys (the multi-key warning); it fires only at boot, never at mutation time.

## Trigger
Peers `10.0.0.1 AS65001` (key A) and `10.0.0.1 AS65002` (key B); `DELETE /api/peers/{AS65002-id}` → new connections of AS65001 are accepted unsigned.

## Fix shape
After delete/clear, re-arm from the remaining rows for that IP: read `GetPeerMd5CredentialsAsync()` filtered to the IP — if rows remain, set the surviving key (preserving the multi-key warning behavior); call with `null` only when no rows remain. PATCH: same, after the store update.

## Regression test
Two peers on one IP with different configured keys; delete one; assert the listener still carries the sibling's key (spy on `ISessionManager.SetPeerMd5Key` / assert via the `ApplyMd5` path).

## Severity
HIGH — silent security-posture downgrade (TCP-MD5 is optional per peer, but a silent downgrade contradicts the operator's configured intent).

## Related
- #36 (per-peer TCP-MD5)

### LINKED ISSUE #419: providers: SSRF blocklist misses the RFC 8215 NAT64 local-use prefix 64:ff9b:1::/48 (literal 64:ff8:1::/96 is wrong)
## Found by
External deep audit (2026-09-03) — SSRF review. Prefix verified against the RFC text (fact-check).

## Summary
The SSRF blocklist intends to block the NAT64 local-use translation prefix, but the literal is wrong twice: `64:ff8:1::/96` instead of **`64:ff9b:1::/48`** (RFC 8215 — the local-use well-known prefix; RFC 6052 §2.1 defines only the well-known `64:ff9b::/96`, which IS correctly blocked at `PrefixSourceUrlValidator.cs:40`).

## Root cause
`BGPLite.Providers/PrefixSourceUrlValidator.cs:41`:
```csharp
IPNetwork.Parse("64:ff8:1::/96"),       // NAT64 well-known, local scope (RFC 6052 §2.1 alternate)
```
- hex wrong: `ff8` ≠ `ff9b`;
- length wrong: `/96` ≠ `/48`;
- citation wrong: the local-use prefix is reserved by RFC 8215, not RFC 6052 §2.1.

An address under `64:ff9b:1::/48` (e.g. `64:ff9b:1::a9fe:1694` embedding 169.254.169.254) matches neither line 40 nor line 41, passes `IsBlockedAddress` at BOTH layers (`ValidateUrlAsync` :198 and `ConnectCallback` :101), and — on any network with a NAT64/CLAT gateway honoring local-use prefixes — the peer-supplied URL fetch reaches the embedded internal IPv4 target. That is exactly the embedding class lines 34-40 exist to close (#158).

## Fix shape
Replace the entry with `IPNetwork.Parse("64:ff9b:1::/48")` and fix the comment to cite RFC 8215. Make the two NAT64 literals table-driven so they cannot drift again.

## Regression test
`IsBlockedAddress(IPAddress.Parse("64:ff9b:1::a9fe:1694")) == true`; keep `64:ff9b::a9fe:1694` blocked (well-known form, already covered).

## Severity
HIGH (conditional — exploitable only on networks that route NAT64 local-use prefixes, but that is precisely an internal-network reachability bypass via peer-controlled input).

## Related
- #158 (embedded-IPv4 SSRF forms), #321 (DNS64 well-known prefix)

### LINKED ISSUE #420: protocol,server: ROUTE_REFRESH for AFI=2 (IPv6/Unicast) is ignored — MP-v6 peers cannot re-request routes
## Found by
External deep audit (2026-09-03) — IPv6 epic #14 follow-up.

## Summary
`BgpSession` handles ROUTE_REFRESH only for AFI=1/SAFI=1 and ignores every other AFI/SAFI with a Debug log (`BGPLite.Server/BgpSession.cs:818-821`). BGPLite advertises MP IPv6/Unicast (RFC 4760 §8), so a conformant peer MAY send ROUTE_REFRESH with AFI=2 for the v6 family (RFC 2918 §2 — the receiving speaker re-sends Adj-RIB-Out for the requested AFI/SAFI). Today such a request is dropped: an IPv6 peer that lost our routes (or wants a re-announce after a policy change) can never get them re-advertised without bouncing the whole session.

## Current behavior
- AFI=1/SAFI=1 → debounced full re-announce (fire-and-forget `RefreshRoutesAsync`, #253).
- AFI=2/SAFI=1 → "RouteRefresh ignored: unsupported AFI/SAFI" at Debug; nothing is sent.

## Required behavior
For an MP-IPv6-negotiated session, an AFI=2/SAFI=1 ROUTE_REFRESH should trigger the same debounced re-announce (the v6 half rides MP_REACH, #14 phase 4). Share the existing per-session rate limit (`MinRouteRefreshInterval`); refreshing both families on an AFI=1 request is also acceptable since the current dump is unified.

## Regression test
Session-level: established session with `_peerMpIpv6Unicast`, scripted ROUTE_REFRESH(2,1) frame → assert an MP_REACH re-advertisement is emitted (and the debounce still collapses a burst).

## Severity
MEDIUM — functional gap on the v6 data path (ignoring is not an RFC violation per se, but it strands conformant v6 peers).

## Related
- #14 (IPv6 epic), RFC 2918 receiving-speaker semantics

### LINKED ISSUE #421: api: peer IP validation is weaker than the YAML path — 0.0.0.0/loopback/multicast accepted as peer rows
## Found by
External deep audit (2026-09-03); found as a parity gap left by #390.

## Summary
The YAML path rejects peer addresses that are `0.0.0.0` (`AppConfig.Validate`, #390), but the management API's `NormalizePeerIp` (`BGPLite.Api/ManagementApi.cs:1498-1509`) accepts any parseable address: `0.0.0.0`, `127.0.0.1`, multicast (`224.0.0.1`), broadcast (`255.255.255.255`) and their IPv6 counterparts (`::`, `ff02::1`) all become peer rows. Such a row can never establish a session, and `SetPeerMd5Key` will even arm a kernel TCP-MD5 entry for a multicast address (`BgpServer.cs:189-197` only validates the password length).

Server validation is authoritative (AGENTS.md) — it currently disagrees with itself across the two peer-creation paths.

## Fix shape
Share the same predicate as the YAML path: unicast only, explicitly reject the unspecified address, loopback, multicast and reserved/broadcast with a clear 400. Auto-registered peers are unaffected (they come from real accepted connections, not this parser).

## Regression test
`PeerInputValidationTests`: POST /api/peers with `0.0.0.0` / `127.0.0.1` / `224.0.0.1` / `255.255.255.255` / `::` / `ff02::1` → 400.

## Severity
MEDIUM (junk rows + kernel socket options on bogus addresses; no session impact).

## Related
- #390, #255

### LINKED ISSUE #422: api,server: peer delete with NULL Asn terminates nothing — the session keeps advertising a deleted peer
## Found by
External deep audit (2026-09-03) — peer lifecycle review.

## Summary
DELETE resolves the session identity as `peer.Asn ?? 0` (`BGPLite.Api/ManagementApi.cs:1100` for `TerminatePeerAsync`; `:1081` for `RefreshPeerAsync`). No live session ever has `RemoteAsn == 0` (AS 0 OPENs are rejected, #300), so for a peer row with NULL Asn the teardown is a silent no-op: the row is deleted while the established session keeps running and advertising — exactly the outcome #323 exists to prevent. `RefreshPeer` then logs a confusing "no session for {Ip} AS0".

## Trigger
Legacy rows from the Ip-only era (pre-#19) or any row written without an ASN. The HTTP API writes Asn today, so this is a legacy-data edge — but the failure is fully silent.

## Fix shape
Pick one and implement:
- terminate by IP when Asn is null (IP-only session matching, with a warning log), or
- refuse deleting a NULL-Asn row while a session for that IP is established (409 with guidance), or
- at minimum, log a warning: "peer row has no ASN; the live session for {Ip} is left running".

The no-op must stop being silent either way.

## Regression test
`PeerDeleteTeardownTests`: seeded NULL-Asn row + established scripted session → DELETE → the session is disposed/its routes withdrawn (or the refusal is returned), and a warning is logged.

## Severity
MEDIUM (legacy-data edge, silent policy violation).

## Related
- #323, #200, #19

### LINKED ISSUE #423: api: rate-limiter partitions are never evicted; with TrustXRealIp the bucket key is client-controlled
## Found by
External deep audit (2026-09-03) — management API review.

## Summary
`CreateRateLimiter` (`BGPLite.Api/ManagementApi.cs:1629-1641`) builds a `PartitionedRateLimiter` with a lazily-created `TokenBucketRateLimiter` per distinct client-IP key. `System.Threading.RateLimiting` has no idle-partition eviction: every IP ever seen (even one rejected request) holds its partition and its AutoReplenishment timer for the process lifetime — unbounded growth on a long-lived deployment (IPv6 /128 clients make it worse).

Sharper variant: with `TrustXRealIp: true` and a pass-through proxy, the resolved key comes straight from the attacker-supplied `X-Real-IP` header (`ResolveClientIp`, `ManagementApi.cs:1672+`) — rotating the header per request (a) defeats the rate limit entirely (fresh bucket each time) and (b) grows the partition table without bound. The spoofing half is documented as an opt-in tradeoff (`AppConfig.cs:58-66`); the missing eviction is documented nowhere.

## Fix shape
- Periodic idle-partition eviction: keep a last-seen stamp per key and park/dispose limiters for idle keys (the existing retired-limiter drain machinery, `ManagementApi.cs:152-158` / `:289`, is a template), or bound the partition count with an LRU.
- Document the TrustXRealIp partition-growth caveat next to the existing spoofing warning.

## Regression test
Unit: drive N distinct keys through the limiter, advance time past an idle window, assert partitions are reclaimed (or a documented bound holds).

## Severity
MEDIUM (memory growth + rate-limit bypass under a documented-but-fragile opt-in).

## Related
- #116, #256

### LINKED ISSUE #424: api: cold-cache /api/asn-lists and /api/peers/{id}/prefixes can occupy the entire in-flight cap
## Found by
External deep audit (2026-09-03) — management API review.

## Summary
GET `/api/asn-lists` (`ManagementApi.cs:1298+`, sequential `GetPrefixCountAsync` per ASN per list) and GET `/api/peers/{id}/prefixes` (`:534`, `:1240`, `CollectPeerPrefixes` `:1250-1290`) perform external fetches (RIPEstat: per-attempt timeout 180s × retries) bounded only by `_shutdownCts`. The API's only always-on bound is the global in-flight cap of 64 (`DefaultInflightCap`, `ManagementApi.cs:61`); the per-IP rate limiter and the global concurrency limiter are both opt-in (off by default). A single unauthenticated client opening 64 such GETs pins the entire in-flight budget for minutes and starves `/api/me` and every mutating route.

## Fix shape
- Add a per-request wall-clock bound (e.g. a 30s linked CTS) on endpoints that can trigger external fetches; return partial results with a "degraded" marker instead of full-table waits, or
- give fetch-bearing GETs a small dedicated semaphore (e.g. 4) so they cannot monopolize the in-flight cap.

## Regression test
Endpoint test with a parked `IPrefixService`: N+1 concurrent fetch-bearing GETs; assert the (N+1)th is rejected/timed out quickly and `/api/server` still answers.

## Severity
MEDIUM (unauthenticated control plane — default bind is loopback, so this needs local or reverse-proxy-reachable access).

## Related
- #119 (in-flight cap), #116 (opt-in limiters)

### LINKED ISSUE #425: providers: one shared circuit breaker couples peer-supplied URLs to operator prefix sources
## Found by
External deep audit (2026-09-03) — resilience review.

## Summary
A single resilience pipeline (retry + circuit breaker: 10 failures / 0.5 ratio / 10s sampling / 30s open) sits on the shared `"http"` named client (`BGPLite/Program.cs:98-120`, `:302-321`). That client serves BOTH operator-configured `PrefixSources` and peer-supplied user URLs (`HttpPrefixProvider` / `PrefixService.GetUserSourcePrefixesAsync`). A handful of peers pointing user sources at blackholed IPs generate repeated failures; the shared breaker opens and every operator source fetch fails fast with `BrokenCircuitException` for 30s windows — fleet-wide suppression of source updates driven entirely by peer-supplied input.

## Fix shape
Either separate named clients (operator sources vs. user URLs), or per-request strategies so peer-URL failures cannot trip the breaker that gates operator fetches. `HttpPrefixProvider` already enforces its own per-fetch budget — the shared breaker mostly adds coupling across a trust boundary, not protection.

## Regression test
Provider-level: open the breaker with user-URL failures, then assert an operator source fetch still performs a real HTTP attempt.

## Severity
MEDIUM (external-input-driven degradation of operator-facing functionality).

## Related
- #107 (resilience pipeline introduction), #320 (user-source fetch budget)

### LINKED ISSUE #426: providers: UserSourceCache/RipeStatPrefixCache bound memory by entry count only — no byte budget, no idle sweep
## Found by
External deep audit (2026-09-03) — memory-bounding review.

## Summary
`UserSourceCache` (1024 entries, `BGPLite.Providers/UserSourceCache.cs:50`) and `RipeStatPrefixCache` (4096 entries, `RipeStatPrefixCache.cs:71`) bound memory by ENTRY COUNT only. Each entry holds a parsed prefix list from a response capped at 10 MB (`MaxResponseBytes`, `HttpPrefixProvider.cs:31` / `RipeStatProvider.cs:16`) — up to ~1M `IpPrefix` (~24B+ each) per entry, so tens of GB worst case. Eviction runs only on insert at capacity (`UserSourceCache.cs:138-168`; `RipeStatPrefixCache` likewise) — expired entries are pinned until the cap is hit, so steady-state memory is "everything fetched recently", not "live entries".

Peer-triggered URL churn (many distinct user-source URLs across peers) is the realistic path to high entry counts; RIPEstat ASNs are operator-bounded in practice.

## Fix shape
- Add byte-budget accounting (sum of parsed list sizes) with eviction of largest/oldest over budget; and/or
- a periodic sweep that drops expired entries even below the cap; and/or
- lower caps for user sources (peer-controlled) vs RIPEstat (operator-controlled) — the trust boundary justifies different limits.

## Regression test
Cache test with `maxEntries=2` and oversized fake lists: assert memory-bound eviction (or a documented cap) holds, and that expired-but-under-cap entries are eventually swept.

## Severity
MEDIUM (memory exhaustion requires adversarial-ish peer behavior; no correctness impact).

## Related
- #165, #261 (the bounding work this extends)

### LINKED ISSUE #427: server,protocol: malformed OPEN in Established is kept as a body error instead of FSM error (5/0)
## Found by
External deep audit (2026-09-03) — protocol conformance sweep.

## Summary
A well-formed OPEN in Established correctly hits the FSM-error default (NOTIFICATION 5/0, `BgpSession.cs:851-864`). A MALFORMED OPEN does not: `ParseOpen` throws `BgpParseException` carrying `OpenMessageError` BEFORE the message-type switch, and the read loop's body-error filter (`BgpSession.cs:732-769`, `when ex.ErrorCode is not null`) treats it as a per-message content error — the session stays up with a Warning. Per RFC 4271 §8.2.2 an OPEN received in Established is an FSM error (NOTIFICATION Finite State Machine Error, release resources, Idle) regardless of body validity; a conformant speaker never even parses the body in that state.

D17 (`docs/DESIGN_DECISIONS.md:180-201`) records the "UPDATE body errors keep the session" deviation; it does not cover OPEN, and the current behavior looks like an accident of dispatch order rather than a recorded decision.

## Fix shape
Either (a) route `type == Open` in Established to the FSM-error branch before body parsing, or (b) if keeping the session is deliberate, record it in the D-catalog next to D17 with the same "a route server must not lose a session over one bad message" rationale.

## Regression test
Session-level: established session, scripted OPEN with a truncated body → assert NOTIFICATION 5/0 + Idle (option a), or pin the keep-up behavior with a comment (option b).

## Severity
LOW (protocol purity; no route or availability impact).

### LINKED ISSUE #428: server: accept loop spins without backoff on persistent AcceptAsync failures
## Found by
External deep audit (2026-09-03) — robustness review.

## Summary
`AcceptLoopAsync` catches all accept failures and continues (`BGPLite.Server/BgpServer.cs:362-368`). A persistent `AcceptAsync` failure — fd exhaustion (EMFILE) is the classic — makes the loop spin at full speed: log Error + immediately retry, thousands of times per second, exactly when the system is already out of resources. The `Volatile.Read(ref _acceptingConnections) == 0` check breaks the loop only on shutdown, not on listener faults.

## Fix shape
A small fixed backoff before retrying non-cancellation accept failures (e.g. 100ms–1s via `Task.Delay` on the loop's token); optionally escalate to recreating the listener socket after N consecutive failures.

## Regression test
Extract the loop's retry policy (delay-after-failure) and assert it directly, or a scripted-socket test asserting the error-log rate is bounded under a persistent accept failure.

## Severity
LOW (self-DoS under pre-existing resource exhaustion; BGP listener only).

### LINKED ISSUE #429: perf(server,providers): quadratic custom-prefix suppression + per-call ToContract allocations on the outbound build
## Found by
External deep audit (2026-09-03) — performance review of the outbound build path.

Two hot-path items; both run per peer per establish/refresh, and `RefreshAllEstablishedAsync` multiplies the cost by fleet size.

### 1. Quadratic custom-prefix suppression
`SuppressCoveredByCustomPrefixes` (`BGPLite.Server/RouteAssembler.cs:310-331`) runs `customRanges.Contains(...)` (List, O(C)) and `customRanges.Any(cr => ...)` inside the per-route `Where` — O(routes × C) UInt128 mask+compares. 11k routes × 1000 custom prefixes ≈ 11M compares per peer per refresh.
Fix: precompute a per-length dictionary (length → set of masked networks) for the coverage check; keep the exact-match fast path via a HashSet.

### 2. Per-call `ToContract` allocation on warm cache hits
`PrefixService.ToContract` (`BGPLite.Providers/PrefixService.cs:243-244`) allocates a fresh `(UInt128, byte, bool)` list on EVERY call of `GetPrefixesAsync` / `GetSourcePrefixesAsync` / `GetUserSourcePrefixesAsync` — including warm hits. `RouteAssembler` invokes these per source per peer (`RouteAssembler.cs:117-122`, `:165-167`, `:369-371`): a fleet refresh of N peers against an 11k-prefix source allocates N×11k tuples per cycle.
Fix: cache the contract-shaped list alongside the native one (invalidate together), or return one shared immutable instance.

Non-goals: LINQ/allocations elsewhere on the path are fine; these two are the ones whose cost scales with route count × peers.

## Severity
LOW–MEDIUM (throughput/memory only; no correctness impact).

## Related
- #305 (same class of fix for attribute-key normalization), #85, #343

### LINKED ISSUE #430: protocol: >4096-byte outbound UPDATE throws after _advertisedPrefixes already recorded the batch
## Found by
External deep audit (2026-09-03) — protocol/robustness sweep.

## Summary
`BgpMessageWriter.WriteHeader` throws `ArgumentOutOfRangeException` when a composed UPDATE exceeds `MaxMessageSize` (`BGPLite.Protocol/BgpMessageWriter.cs:33-45`). In `BgpSession.SendRoutesAsync` routes are appended to `_advertisedPrefixes` BEFORE their batch is sent (`BgpSession.cs:1278-1288`): when `WriteUpdate` throws mid-send, some routes are already recorded as advertised but never reached the wire. The mirror then diverges from reality — a later `WithdrawAllAsync` sends spurious withdrawals for never-announced prefixes (harmless but noisy), and `_advertisedCount` over-reports.

Reaching >4096 bytes requires a community group large enough (e.g. hundreds of communities on one community set — constructible from a peer-supplied COMMUNITY attribute that is re-advertised outbound).

## Fix shape
Build-then-commit: append a batch to `_advertisedPrefixes` only after its UPDATE serialized+sent successfully, or pre-validate the composed size against `MaxMessageSize` and split the batch. Also log the >4096 case loudly — it indicates an outbound construction bug, not peer input.

## Regression test
`SendRoutesAsync` with a route carrying enough communities to exceed 4096 bytes in one group → assert the send path fails cleanly and `_advertisedPrefixes` contains only actually-sent routes (or the batch is split and everything goes out).

## Severity
LOW.

### LINKED ISSUE #431: api: DbUpdateException mapping answers 409 for NOT NULL/CHECK violations
## Found by
External deep audit (2026-09-03) — API error-mapping review.

## Summary
`MapExceptionToResponse` (`BGPLite.Api/ManagementApi.cs:497-505`) answers 409 "The resource already exists or conflicts with the current state" for EVERY `DbUpdateException` whose extended SQLite code is not 787. That bucket includes NOT NULL (1299) and CHECK (275) violations — schema/data bugs, not client conflicts. API consumers debugging a 409 for a create/update that could not possibly conflict are misled; the real cause is only in the server log.

## Fix shape
Map by extended code: 787 → 404 (existing), 2067 → 409 (existing), anything else → 500 with the generic message. One extra branch.

## Regression test
`ExceptionMappingTests`: `DbUpdateException` wrapping `SqliteException` with extended code 275 / 1299 → 500; 2067 → 409; 787 → 404.

## Severity
LOW.

ruhex:
Closes #416
Closes #417
Closes #418
Closes #419
Closes #420
Closes #421
Closes #422
Closes #423
Closes #424
Closes #425
Closes #426
Closes #427
Closes #428
Closes #429
Closes #430
Closes #431

## Summary
16 issues from the 2026-09-03 external audit (1 critical, 3 high, 7 medium, 5 low), one squash-merged PR each into `dev`, each verified individually (full suite + CodeQL 0 alerts on its merge-ref).

### Critical / High
- **#416 / #433** — providers: `_ruGate` deadlock. The RU load is callback-free (`LoadDefaultAsync`) and the convergence push fires after `_ruGate.Release()` — a changed default source no longer permanently stalls every RU consumer. Red/green deadlock test.
- **#417 / #434** — providers: a failed default-source load (including the 30s negative-backoff path) surfaces as a failure — stale-served or thrown — instead of being cached as an empty RU set for 1h. Unused swallowing `GetDefaultAsync` removed.
- **#418 / #435** — api,server: peer delete/PATCH re-arms the per-IP TCP-MD5 key from SURVIVING rows (shared resolver with the startup bootstrap) — deleting one sibling no longer silently disarms RFC 2385 for the others.
- **#419 / #436** — providers: SSRF blocklist literal fixed to the RFC 8215 NAT64 local-use prefix `64:ff9b:1::/48` (was `64:ff8:1::/96`).

### Medium
- **#420 / #437** — server: ROUTE_REFRESH for AFI=2 honored on MP-IPv6 sessions (RFC 2918 §2); v6 peers can re-request routes.
- **#421 / #440** — api: `NormalizePeerIp` rejects unspecified/loopback/multicast/broadcast (both families, mapped forms normalized) — parity with the YAML path's #390 validation.
- **#422 / #439** — api,server: NULL-Asn peer delete terminates its sessions by IP (`ISessionManager.TerminatePeerByIpAsync`) instead of a silent no-op teardown.
- **#423 / #446** — api: `ClientIpRateLimiter` replaces `PartitionedRateLimiter` — idle IP partitions (and their replenishment timers) are evicted; TrustXRealIp key churn no longer grows the table without bound.
- **#424 / #443** — api: 30s wall-clock budget (shutdown-linked) on `/api/asn-lists` and `/api/peers/{id}/prefixes`; expiry serves partial results instead of pinning the in-flight cap.
- **#425 / #444** — providers: peer-supplied user sources fetch on their own named client (retry-only) — their failures can no longer open the shared circuit breaker gating operator sources.
- **#426 / #445** — providers: amortized expired-entry sweeps in both caches + a 2M-prefix budget on `UserSourceCache` — expired entries are no longer pinned until the cap.

### Low
- **#427 / #438** — server: a malformed OPEN in Established is an FSM error (5/0) like a well-formed one (RFC 4271 §8.2.2); D17 scope note added.
- **#428 / #442** — server: 500ms backoff on persistent accept failures (no more hot spin under fd exhaustion).
- **#429 / #447** — perf: custom-prefix suppression via a coverage index (O(routes+customs)); contract projection memoized per native list (ConditionalWeakTable).
- **#430 / #448** — server: the send mirror records a batch only after its UPDATE reaches the wire — no phantom withdrawals after a failed oversize send.
- **#431 / #441** — api: SQLite constraint mapping by extended code (787→404, 2067→409, else 500) — NOT NULL/CHECK no longer mislabeled 409.

## Verification (per PR, and re-run on the dev head)
- Every PR: dotnet format → full `dotnet build` (0 errors) → full `dotnet test` (green; 1010 → 1045 tests, all growth = new regression tests) → CI (build-test, format-check, analyze, CodeQL) → CodeQL alerts on merge-ref: 0.
- Every behavioral fix carries a red/green regression test (fails against the pre-fix code) where deterministically reproducible; two cases (accept-loop backoff, rate-limiter sweep races) document why not.
- Dev head re-verified before this PR: build 0 errors, **1045/1045 tests green**.
- CodeRabbit was disabled on dev-based PRs by `.coderabbit.yaml` — its full review happens HERE; findings will be addressed per-finding (fix or reasoned reply) before merge.

## RFC
- #420: RFC 2918 §2; #427: RFC 4271 §8.2.2 (+ D17 scope note in docs/DESIGN_DECISIONS.md); #419: RFC 6052 §2.1 / RFC 8215; #418: RFC 2385 enforcement consistency.

## Risk notes (behavior changes vs main)
- RU failure path now degrades (stale/throw) instead of advertising an empty set for 1h.
- NULL-Asn deletes actually stop sessions (by IP, with a warning); non-unicast peer IPs return 400 at the API.
- Peer-URL failures no longer suppress operator sources (separate retry-only pipeline).
- Rate-limiter buckets reset after the idle threshold (default ≥5 min) — a returning IP starts refilled.
- Oversize UPDATE sends fail loudly without corrupting the send mirror.

## Summary by CodeRabbit

* **New Features**
  * Added support for IPv6 peer addresses while rejecting unusable addresses.
  * Added bounded external fetches that return partial results with warnings when providers time out.
  * Added safer handling for legacy peers and shared TCP-MD5 keys.
  * Added dedicated handling for peer-supplied prefix sources.

* **Bug Fixes**
  * Improved rate-limit and prefix-cache cleanup to remove idle or expired entries.
  * Prevented failed route advertisements from causing incorrect withdrawals.
  * Improved BGP Route Refresh support for negotiated IPv6 routes.
  * Corrected error responses for database conflicts and missing records.
  * Hardened SSRF protection for NAT64 address ranges.

### DIFF
--- a/BGPLite.Api/ClientIpRateLimiter.cs
+++ b/BGPLite.Api/ClientIpRateLimiter.cs
@@ -1,0 +1,135 @@
+using System.Collections.Concurrent;
+using System.Threading.RateLimiting;
+using BGPLite.Configuration;
+
+namespace BGPLite.Api;
+
+/// <summary>
+/// Per-client-IP token-bucket rate limiter with IDLE-PARTITION EVICTION (#423).
+/// <para>
+/// System.Threading.RateLimiting's <c>PartitionedRateLimiter</c> has no idle eviction: every
+/// client IP ever seen held its bucket (and its AutoReplenishment timer) for the process lifetime —
+/// unbounded growth on a long-lived deployment, and with <c>Api.TrustXRealIp</c> the partition key
+/// is client-controlled (rotating <c>X-Real-IP</c> minted fresh buckets per request, both defeating
+/// the limit and growing the partition table). This registry keeps one
+/// <see cref="TokenBucketRateLimiter"/> per IP, stamps last-access on every acquire, and —
+/// amortized, every <see cref="SweepEvery"/> acquires — removes and disposes partitions idle for
+/// longer than the threshold (default: 20 replenishment periods, floor 5 minutes).
+/// </para>
+/// <para>
+/// A request racing a sweep's dispose gets a fresh bucket (the acquire retries once) — a rare
+/// request loses its refill state, never correctness. 429/deny semantics, queue-free, are
+/// identical to the <c>PartitionedRateLimiter</c> this replaces (#116).
+/// </para>
+/// </summary>
+internal sealed class ClientIpRateLimiter : IDisposable, IAsyncDisposable
+{
+    internal const int DefaultSweepEvery = 64;
+    private static readonly TimeSpan MinIdleThreshold = TimeSpan.FromMinutes(5);
+
+    private readonly ConcurrentDictionary<string, (TokenBucketRateLimiter Limiter, long LastAccessTicks)> _partitions = new();
+    private readonly TokenBucketRateLimiterOptions _options;
+    private readonly TimeProvider _timeProvider;
+    private readonly TimeSpan _idleThreshold;
+    private readonly int _sweepEvery;
+    private int _acquiresSinceSweep;
+
+    public ClientIpRateLimiter(
+        ApiRateLimitConfig cfg,
+        TimeProvider? timeProvider = null,
+        TimeSpan? idleThreshold = null,
+        int? sweepEvery = null)
+    {
+        _options = new TokenBucketRateLimiterOptions
+        {
+            TokenLimit = Math.Max(1, cfg.TokenLimit),
+            TokensPerPeriod = Math.Max(1, cfg.TokensPerPeriod),
+            ReplenishmentPeriod = TimeSpan.FromSeconds(Math.Max(1, cfg.PeriodSeconds)),
+            QueueLimit = 0,         // deny immediately (429) when no tokens — never queue
+            AutoReplenishment = true
+        };
+        _timeProvider = timeProvider ?? TimeProvider.System;
+        _idleThreshold = idleThreshold ?? TimeSpan.FromTicks(Math.Max(MinIdleThreshold.Ticks, 20 * _options.ReplenishmentPeriod.Ticks));
+        _sweepEvery = sweepEvery ?? DefaultSweepEvery;
+    }
+
+    /// <summary>Partitions currently tracked (test/observability — eviction coverage).</summary>
+    internal int TrackedCount => _partitions.Count;
+
+    public ValueTask<RateLimitLease> AcquireAsync(string clientIp)
+    {
+        // #423: amortized idle-partition eviction — see the class doc.
+        if (Interlocked.Increment(ref _acquiresSinceSweep) >= _sweepEvery)
+        {
+            Volatile.Write(ref _acquiresSinceSweep, 0);
+            SweepIdle();
+        }
+
+        var limiter = GetOrAdd(clientIp);
+        RateLimitLease lease;
+        try
+        {
+            lease = limiter.AttemptAcquire();
+        }
+        catch (ObjectDisposedException)
+        {
+            // Lost a race with the eviction sweep's Dispose: mint a fresh bucket and retry once.
+            // Narrow and benign — the request sees an unfilled bucket instead of failing.
+            limiter = new TokenBucketRateLimiter(_options);
+            _partitions[clientIp] = (limiter, Stamp());
+            lease = limiter.AttemptAcquire();
+        }
+
+        return ValueTask.FromResult(lease);
+    }
+
+    private TokenBucketRateLimiter GetOrAdd(string clientIp)
+    {
+        var now = Stamp();
+        while (true)
+        {
+            if (_partitions.TryGetValue(clientIp, out var existing))
+            {
+                // Touch last-access; the limiter reference is unchanged, so the tuple write is
+                // safe even against a concurrent sweep (compare-remove either wins before the
+                // touch — the request retries on a fresh bucket — or loses and the touch keeps
+                // the partition alive).
+                _partitions[clientIp] = (existing.Limiter, now);
+                return existing.Limiter;
+            }
+
+            var created = new TokenBucketRateLimiter(_options);
+            if (_partitions.TryAdd(clientIp, (created, now)))
+                return created;
+            created.Dispose(); // lost the first-insert race — dispose the spare (no timer leak)
+        }
+    }
+
+    private void SweepIdle()
+    {
+        var now = Stamp();
+        foreach (var (key, entry) in _partitions)
+        {
+            if (now - entry.LastAccessTicks < _idleThreshold.Ticks) continue;
+            // Compare-remove: an entry touched since the snapshot is kept (the pair no longer
+            // matches), so an active partition is never disposed mid-request.
+            if (_partitions.TryRemove(new KeyValuePair<string, (TokenBucketRateLimiter, long)>(key, entry)))
+                entry.Limiter.Dispose();
+        }
+    }
+
+    private long Stamp() => _timeProvider.GetUtcNow().UtcDateTime.Ticks;
+
+    public void Dispose()
+    {
+        foreach (var (_, entry) in _partitions)
+            entry.Limiter.Dispose();
+        _partitions.Clear();
+    }
+
+    public ValueTask DisposeAsync()
+    {
+        Dispose();
+        return ValueTask.CompletedTask;
+    }
+}
--- a/BGPLite.Api/ClientIpRateLimiter.cs
+++ b/BGPLite.Api/ClientIpRateLimiter.cs
@@ -73,10 +73,10 @@ internal sealed class ClientIpRateLimiter : IDisposable, IAsyncDisposable
         }
         catch (ObjectDisposedException)
         {
-            // Lost a race with the eviction sweep's Dispose: mint a fresh bucket and retry once.
-            // Narrow and benign — the request sees an unfilled bucket instead of failing.
-            limiter = new TokenBucketRateLimiter(_options);
-            _partitions[clientIp] = (limiter, Stamp());
+            // Lost a race with the eviction sweep's Dispose (CodeRabbit on #450): the sweep
+            // removed the dead partition, so re-read via GetOrAdd — this mints a fresh bucket
+            // WITHOUT blindly overwriting whatever a concurrent request may have installed.
+            limiter = GetOrAdd(clientIp);
             lease = limiter.AttemptAcquire();
         }
 
@@ -90,12 +90,13 @@ internal sealed class ClientIpRateLimiter : IDisposable, IAsyncDisposable
         {
             if (_partitions.TryGetValue(clientIp, out var existing))
             {
-                // Touch last-access; the limiter reference is unchanged, so the tuple write is
-                // safe even against a concurrent sweep (compare-remove either wins before the
-                // touch — the request retries on a fresh bucket — or loses and the touch keeps
-                // the partition alive).
-                _partitions[clientIp] = (existing.Limiter, now);
-                return existing.Limiter;
+                // Touch last-access CONDITIONALLY (CodeRabbit on #450): a sweep may have removed
+                // and disposed this partition between the read and the write — a blind indexer
+                // write would resurrect the disposed limiter. TryUpdate succeeds only against
+                // the exact tuple read; on failure the loop re-reads (or mints a fresh bucket).
+                if (_partitions.TryUpdate(clientIp, (existing.Limiter, now), existing))
+                    return existing.Limiter;
+                continue;
             }
 
             var created = new TokenBucketRateLimiter(_options);
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1555,8 +1555,11 @@ public sealed class ManagementApi : IHostedService, IDisposable
     /// <c>ToString()</c> collapses all of them onto the form the BGP path will look for.
     /// </para>
     /// <para>
-    /// IPv6 is rejected rather than mapped: BGPLite is IPv4-unicast only (#14 tracks the rest), so
-    /// <c>::ffff:1.2.3.4</c> would produce a peer no session can match.
+    /// Both families are accepted (#14 phase 5): a peer may be an IPv6 host. Addresses no BGP
+    /// session can ever originate from — unspecified, loopback, multicast, and the IPv4 broadcast
+    /// address — are rejected (#421), the API-side parity of the YAML path's fail-loud validation
+    /// (#390): such rows can never match a session, and a multicast row would even arm a kernel
+    /// TCP-MD5 entry.
     /// </para>
     /// </summary>
     internal static string? NormalizePeerIp(string? ip)
@@ -1567,9 +1570,29 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // plain IPv4 form so the row matches the address form the dual-mode listener reports.
         if (address.IsIPv4MappedToIPv6)
             address = address.MapToIPv4();
-        // Both families are accepted (#14 phase 5): a peer may be an IPv6 host; the canonical
-        // ToString form matches what the accept loop stores.
+        if (IsNonPeerAddress(address))
+            return null;
+        // The canonical ToString form matches what the accept loop stores.
         return address.ToString();
+    }
+
+    /// <summary>
+    /// The unspecified, loopback, multicast and (IPv4) broadcast addresses — unusable as a peer
+    /// identity. internal static for unit tests.
+    /// </summary>
+    internal static bool IsNonPeerAddress(IPAddress address)
+    {
+        if (address.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork)
+        {
+            var b = address.GetAddressBytes();
+            return b[0] == 0                                                 // 0.0.0.0/8 — unspecified / current-network
+                || b[0] == 127                                               // 127/8 — loopback
+                || (b[0] & 0xF0) == 0xE0                                     // 224/4 — multicast
+                || (b[0] == 255 && b[1] == 255 && b[2] == 255 && b[3] == 255); // broadcast
+        }
+        return IPAddress.IPv6Any.Equals(address)
+            || IPAddress.IPv6Loopback.Equals(address)
+            || address.IsIPv6Multicast;
     }
 
     /// <summary>
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1107,9 +1107,29 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         _logger.LogInformation("Updated peer {Id}", SanitizeForLog(peerId));
 
-        _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn ?? 0);
+        RequestPeerRefresh(peerId, peer);
 
         return await HandleGetPeer(peerId);
+    }
+
+    /// <summary>
+    /// Fire-and-forget per-peer refresh after a config change. A NULL-Asn row (legacy Ip-only era)
+    /// cannot match any live session by (Ip, Asn) — no session ever has RemoteAsn 0 (#300) — so
+    /// the refresh is skipped with a warning instead of the confusing "no session for {Ip} AS0"
+    /// the old <c>asn ?? 0</c> produced (#422).
+    /// </summary>
+    private void RequestPeerRefresh(string peerId, Peer peer)
+    {
+        if (peer.Asn.HasValue)
+        {
+            _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn.Value);
+        }
+        else
+        {
+            _logger.LogWarning(
+                "Peer row {Id} at {Ip} has NULL Asn — refresh skipped (no (Ip, Asn) session match is possible)",
+                SanitizeForLog(peerId), peer.Ip);
+        }
     }
 
     internal async Task<ApiResponse> HandleDeletePeer(string peerId)
@@ -1126,7 +1146,21 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // send-timeout backstop) from pinning the DELETE request; the session is disposed even
         // when the Cease send is cancelled.
         using var bound = new CancellationTokenSource(TimeSpan.FromSeconds(10));
-        await _sessionManager.TerminatePeerAsync(peer.Ip, peer.Asn ?? 0, bound.Token);
+        if (peer.Asn.HasValue)
+        {
+            await _sessionManager.TerminatePeerAsync(peer.Ip, peer.Asn.Value, bound.Token);
+        }
+        else
+        {
+            // #422: a NULL-Asn row (legacy Ip-only era) cannot be matched by (Ip, Asn) — no live
+            // session ever has RemoteAsn 0 (AS 0 OPENs are rejected, #300) — so the pre-#422
+            // `asn ?? 0` teardown was a SILENT no-op and the session kept advertising a deleted
+            // peer. Terminate by IP instead, and say so in the log.
+            _logger.LogWarning(
+                "Deleting peer row {Id} at {Ip} with NULL Asn — terminating ALL its sessions by IP",
+                SanitizeForLog(peerId), peer.Ip);
+            await _sessionManager.TerminatePeerByIpAsync(peer.Ip, bound.Token);
+        }
 
         await _store.DeletePeerAsync(peerId);
 
@@ -1211,7 +1245,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         // Trigger refresh so the peer receives the new source's prefixes immediately —
         // same pattern as CreatePeer/UpdatePeer. Pass ASN so shared-IP peers aren't refreshed (#200).
-        _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn ?? 0);
+        RequestPeerRefresh(peerId, peer);
 
         _logger.LogInformation("Added source '{Name}' ({Url}) to peer {PeerId}",
             SanitizeForLog(data.Name), SanitizeForLog(data.Url), SanitizeForLog(peerId));
@@ -1228,7 +1262,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
             return ApiResponse.Error($"Source '{sourceId}' not found", 404);
 
         // Trigger refresh so the source's prefixes are withdrawn immediately (#200: ASN-scoped).
-        _ = _sessionManager.RefreshPeerAsync(peer.Ip, peer.Asn ?? 0);
+        RequestPeerRefresh(peerId, peer);
 
         _logger.LogInformation("Deleted source {SourceId} from peer {PeerId}", SanitizeForLog(sourceId), SanitizeForLog(peerId));
         return ApiResponse.Ok(new { id = sourceId, deleted = true });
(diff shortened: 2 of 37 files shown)


ruhex:
Per-finding verdicts on the CodeRabbit review (each verified against the code before acting):

1. **ClientIpRateLimiter race (Major) — FIXED** in #451 → dev (0d27a8e): conditional TryUpdate touch + GetOrAdd re-read on the OCE retry; a disposed limiter can never be written back and concurrent mints no longer overwrite each other.
2. **UserSourceCache gate lifetime (Major) — FIXED** in #451: per-URL active-loader refcount; neither the sweep nor eviction removes a gate with active loaders (the RipeStatPrefixCache #267-item-3 invariant).
3. **BgpSession mirror granularity (Major) — FIXED** in #451: the mirror is committed per EMITTED UPDATE inside SendUpdateBatchAsync (covers both classic-NLRI and MP_REACH prefixes), so a throwing group neither drags earlier groups' entries down nor records its own.
4. **RouteAssembler coverage lookup (perf, quick) — FIXED** in #451: per-length hash sets with the mask stored once.
5. **V6 refresh test determinism (minor) — FIXED** in #451: the AFI=1 control polls for the initial dump.
6. **PrefixService RU-projection invalidation (Major) — tracked, not patched here**: the asymmetry (auto-refresh-detected default-source changes leave the 1h RU projection stale until its TTL expires) predates this integration — the 1h projection is the #229 design. The proper fix is a source-version stamp in the projection cache, a design change rather than an integration tweak. Tracked as #452.

All checks re-run green on the updated head; CodeQL 0 alerts.

Issue #292, opened 2026-08-26
chore: protocol/server audit findings — NEXT_HOP semantics, AS 0 (RFC 7607), header subcodes, KEEPALIVE length, AS_PATH loop, max-prefix, aggregator allocations, size-aware batching, SQLite CVE


ruhex:
Grouped low/medium findings from the 2026-08-27 protocol/server audit. Each was verified against `main` @ 1224144; split out if any turns out to be larger than it looks.

---

## 1. NEXT_HOP is never semantically validated (RFC 4271 §6.3, subcode 8)

> If the NEXT_HOP attribute field is semantically incorrect ... the Error Subcode MUST be set to Invalid NEXT_HOP Attribute. The determination of semantic correctness ... [includes] that the IP address is not the IP address of the BGP speaker, and is a valid unicast address.

`UpdateCodec.ParseRouteAttributes` only length-checks NEXT_HOP; `AttributeHelper.ReadNextHop` reads the 4 octets unconditionally. All of the following are accepted and installed in the `RouteTable`:

```
NEXT_HOP 0.0.0.0            -> ACCEPTED
NEXT_HOP 127.0.0.1          -> ACCEPTED   (loopback)
NEXT_HOP 224.0.0.1          -> ACCEPTED   (multicast / class D)
NEXT_HOP 255.255.255.255    -> ACCEPTED   (broadcast)
```

**Fix**: reject `0.0.0.0`, loopback (`127/8`), multicast (`224/4`), reserved (`240/4`), and the local router-id, with `InvalidNextHopAttribute` (8) via the existing treat-as-withdraw path. **Severity: MEDIUM.**

---

## 2. AS 0 accepted in OPEN and in AS_PATH (RFC 7607)

RFC 7607 §2:
> A BGP speaker that receives an OPEN message with "My Autonomous System" of 0 ... MUST send a NOTIFICATION ... "Bad Peer AS".
> ... an UPDATE with an AS_PATH attribute that contains AS 0 ... MUST be considered as malformed [treat-as-withdraw].

Confirmed:

```
OPEN My-AS=0, 4-octet capability AS=0   -> OpenNegotiator.Validate ACCEPTED, remoteAsn=0
UPDATE AS_PATH = [100, 0]               -> ParseRouteAttributes ACCEPTED, asPath=[100,0]
```

[...]

PR #464, opened 2026-09-03
chore: integrate dev — audit follow-ups #452–#457

### LINKED ISSUE #452: providers: invalidate the 1h RU projection when the default source changes via the auto-refresh path
## Found by
CodeRabbit review on the integration PR #450; independently verified against the code. Pre-existing asymmetry (predates #416/#433), surfaced by the review.

## Summary
The RU/default projection (`PrefixService._ruCache`) has its own 1h TTL, separate from the source-level cache in `PrefixSourceService`. When the default source changes via the **auto-refresh path** (`PrefixSourceService.RefreshAsync` — deliberately callback-free), the push (`LoopAsync → RefreshAllEstablishedAsync`) rebuilds every session's route set, but each session's `GetRuPrefixesAsync` takes the fresh-cache fast path and re-advertises the **old** RU projection. The updated routes only land when the RU TTL expires (up to 1h later).

The connect-path (GetRuPrefixesAsync → `LoadDefaultAsync`) does invalidate correctly since #416/#417 — `changed` is computed at load time and the projection is written together with the push.

## Required behavior
A default-source change detected by ANY path must invalidate (or version-stamp) the RU projection before the sessions rebuild.

## Fix shape
Version-stamp the projection: `PrefixSourceService` keeps a monotonic per-name (or global) version bumped on every committed cache write; `LoadDefaultAsync` returns it; the RU fast path requires `storedVersion == currentVersion` — a mismatch forces the slow path (which re-projects from the already-new source cache and suppresses the duplicate push because `changed` is false there). Deterministic test: auto-refresh-style refresh of the default source → sessions rebuilt with NEW routes immediately, not after 1h.

## Severity
MEDIUM — bounded staleness (≤1h) of the RU set for unconfigured peers after an auto-refresh-detected change; requires AutoRefresh enabled + DefaultPrefixSource configured.

### LINKED ISSUE #453: server,protocol: malformed OPEN in OpenConfirm answers a body error (2/x) instead of FSM error (5/0)
## Found by
Internal code audit (2026-09-03, post-#432 follow-up). Related to #427 — same message class, different FSM phase that #427's fix did not cover.

## Summary
The OpenConfirm handshake read in `BgpSession.RunAsync` (`BGPLite.Server/BgpSession.cs:404-449`) answers a *parsed* OPEN via the `default` switch arm with Finite State Machine Error 5/0 — correct per RFC 4271 §8.2.2. But an OPEN whose body fails to parse never reaches that switch: `BgpMessageReader.ReadMessage` throws `BgpParseException(OpenMessageError)` inside `ReceiveMessageAsync`, which propagates out of the handshake to `RunAsync`'s catch (`BgpSession.cs:491-505`, the #223 path) and emits NOTIFICATION **Open Message Error (2/x)**.

RFC 4271 §8.2.2: in OpenConfirm, an OPEN is an FSM error regardless of body validity — exactly the reasoning #427 applied to Established (fixed there in the read loop, `BgpSession.cs:732-747`). The handshake path has the same defect with a different subcode.

Impact is subcode-only: the session tears down either way; the peer's operator sees a misleading 2/x diagnostic that suggests the OPEN body was malformed when the real problem is "you already sent OPEN".

## Fix shape
Wrap the OpenConfirm `ReceiveMessageAsync` call with the same treatment as the #427 Established branch: `catch (BgpParseException) when (ex.ErrorCode == OpenMessageError)` → CAS-latch `TeardownReason.LocalCease` → send NOTIFICATION 5/0 → return. One catch block, mirrors `BgpSession.cs:732-747`.

## Severity
LOW — wrong NOTIFICATION subcode only (session teardown is correct in both cases).

## Regression test
Establish a session to OpenConfirm, send a truncated/invalid OPEN body → assert exactly one NOTIFICATION with Error=5 (FSM Error), SubError=0, then Idle. Model on the Established test added for #427 (`MalformedUpdateResilienceTests`).

### LINKED ISSUE #454: api: cold-cache GET /api/as/{asn}/prefixes?count=true misses the #424 external-fetch budget
## Found by
Internal code audit (2026-09-03, post-#432 follow-up). Gap in the #424 fix — the third cold-fetch GET endpoint was not covered.

## Summary
#424 bounded `GET /api/asn-lists` and `GET /api/peers/{id}/prefixes` with `ExternalFetchBudget` (30 s wall clock) because a cold RIPEstat fetch is minutes-scale and pins an in-flight slot (global cap 64). `GET /api/as/{asn}/prefixes?count=true` (`BGPLite.Api/ManagementApi.cs:1516-1542`) still runs `GetPrefixCountAsync` against `_shutdownCts.Token` only — the same minutes-scale fetch holds its slot and the client for the full timeout×retries chain. A few concurrent cold-ASN lookups against a slow upstream degrade every other API route — the exact failure mode #424 describes, on the endpoint it missed.

## Fix shape
Same pattern as the two covered endpoints: linked CTS off `_shutdownCts` + `CancelAfter(ExternalFetchBudget)`; on budget expiry answer a stable non-revealing error and let shutdown still propagate.

## Severity
LOW–MEDIUM — bounded by the 64-slot cap and one fetch chain per request, but it is the same class #424 was filed for.

## Regression test
Handler wired to a hanging message handler (the #424 test shape) → `GET /api/as/65001/prefixes?count=true` answers within `ExternalFetchBudget`, in-flight slot released.

### LINKED ISSUE #455: api,server: peer create arms the shared-IP TCP-MD5 key directly — bypasses the #418 shared resolver
## Found by
Internal code audit (2026-09-03, post-#432 follow-up). Gap in the #418 fix — the create path was not covered.

## Summary
#418 made peer delete/PATCH and the startup bootstrap resolve the per-IP TCP-MD5 key through `ResolveSharedIpKey` (`BGPLite.Api/ManagementApi.cs:273-282`): a deterministic ordinal pick among ALL surviving rows for the IP, with a warning when siblings declare different keys. `HandleCreatePeer` still arms the key directly from the new row — `SetPeerMd5Key(normalizedIp, data.Md5Password)` (`ManagementApi.cs:994-995`). Creating a second peer on a source IP that already carries a different key silently overwrites the armed key (last-writer-wins, no warning), and the armed state can disagree with what a subsequent delete of that row would resolve back to.

TCP keys by address, so this is a consistency/observability drift, not a security hole — but it is exactly the cross-path divergence #418 eliminated for the other three paths.

## Fix shape
After `SavePeerConfigurationAsync`, call `RearmPeerIpMd5KeyAsync(normalizedIp)` instead of the direct `SetPeerMd5Key` — the resolver then applies the same deterministic pick + warning the create path currently bypasses.

## Severity
LOW.

## Regression test
Two peer rows on one IP with different keys → create the second → the armed key equals `ResolveSharedIpKey`'s pick and the siblings-disagree warning is logged.

### LINKED ISSUE #456: docs: record the outbound AS_PATH re-origination (local-ASN-only) decision in DESIGN_DECISIONS
## Found by
Internal code audit (2026-09-03). Documentation-only.

## Summary
Every outbound UPDATE carries an AS_PATH containing ONLY the local ASN: `UpdateCodec.BuildUpdateAttributes`/`BuildAsPathAttributes` build the path from `localAsn` (`BGPLite.Protocol/UpdateCodec.cs:30-70`) and `RouteAssembler.MakeRoute` passes `asPath: null` (`BGPLite.Server/RouteAssembler.cs:360-371`, see also the comment at :119-121). BGPLite never prepends the received path — it re-originates every advertised prefix.

This is a deliberate route-server model, but it is not recorded in `docs/DESIGN_DECISIONS.md` (D1–D22 cover FSM, teardown, GR, aggregation, filters — not path construction). The catalog is actively used to adjudicate "bug vs decision"; an operator or auditor comparing against RFC-strict speakers will otherwise re-derive this and file it. The consequence worth recording explicitly: AS-loop detection at RECEIVING speakers (RFC 4271 §9.1.2) cannot recognize their own prefix coming back, because the returned path carries only BGPLite's ASN. (The inbound half — excluding routes whose AS_PATH contains the local ASN — is implemented; see #292 item 6.)

## Fix shape
New D-entry: "Outbound routes are re-originated (AS_PATH = local ASN only)" — rationale, the loop-detection consequence for peers, and a pointer to the inbound loop-exclusion rule in `BgpSession.HandleUpdateAsync`.

## Severity
LOW (docs-only; no code change required).

### LINKED ISSUE #457: protocol: settle the >4096-byte outbound UPDATE policy — the #292-item-9 size-aware batching is still unimplemented (initial-send teardown remains)
## Found by
Internal code audit (2026-09-03). This is a "fixed twice, root fix still pending" case — recorded so the third report can be the last one.

## History
- **#292 item 9** (2026-08-27): "SendRouteBatchAsync batches by NLRI count only, never by encoded size … a peer configured with enough communities makes the UPDATE exceed `MaxMessageSize` (4096), at which point `WriteHeader` throws `ArgumentOutOfRangeException`. On the initial dump that path reaches `RunAsync`'s generic catch and tears the session down. **Fix: size-aware batching** (split when the encoded length would exceed the cap)."
- **#430** (2026-09-03): re-reported the same defect via its mirror-corruption aspect; fixed by build-then-commit mirror ordering (#448). The #292-item-9 size-aware batching / size pre-validation was **not** implemented.
- Current `main`: `SendRoutesAsync` still bounds batches by `maxNlriPerUpdate = 100` only (`BGPLite.Server/BgpSession.cs:1290`); `BgpMessageWriter.WriteHeader` still throws AOORE past 4096 (`BGPLite.Protocol/BgpMessageWriter.cs:33-45`).

## Current behavior
- Refresh path: contained — `RefreshCycleAsync` logs and the session stays up; the mirror stays consistent (per-UPDATE commit, pinned by `BgpSessionPhantomWithdrawalTests`).
- Initial-send path (`RunAsync` → `SendAllRoutesAsync`, `BgpSession.cs:463`): the AOORE reaches the generic catch → best-effort Cease → session teardown. A peer that just connected gets zero routes and a reset, and reconnects into the same state until the offending route/config changes.

## Reachability (narrower than #430's description)
Outbound communities are config-stamped (one per source — `ConfigCommunityResolver`), large communities never propagate outbound, and peer-learned routes are never re-advertised — so a REMOTE peer cannot construct this. Reaching 4096 requires ~890 communities on one route, i.e. hundreds of configured sources covering one prefix — practically a misconfiguration, not an attack. (#430's "peer-supplied COMMUNITY re-advertised outbound" premise was incorrect; the mirror-ordering defect it fixed was real regardless.)

## Required decision
Either
- **(a)** implement the #292 item 9 fix: pre-validate the composed batch size against `MaxMessageSize` and split (or drop the offending group with a loud Warning) — initial send included; or
- **(b)** record the accepted-risk decision in `docs/DESIGN_DECISIONS.md` (config-only reachability, refresh path contained) and add a Warning log when a send path approaches the cap.

(a) is the honest fix; (b) is acceptable given the reachability analysis. What is not acceptable is a third re-report.

## Severity
LOW (reachability) — but the initial-send failure mode is the worst shape ("fails in the worst possible way", #292 item 9).

## Regression test
Extend `BgpSessionPhantomWithdrawalTests`: poison route present BEFORE the session establishes → assert the initial send behaves per the settled decision — (a) everything except the oversized group is advertised (or the group is split), or (b) documented teardown. Pin whichever is chosen.

ruhex:
Closes #452
Closes #453
Closes #454
Closes #455
Closes #456
Closes #457

Batch of six audit-follow-up fixes, each individually reviewed and merged into `dev` (PRs #458–#463). Full suite **1052 passed / 0 failed**; every fix carries a red/green regression test.

## #452 — providers: invalidate the 1h RU projection on auto-refresh changes (PR #458)
`PrefixService._ruCache` (1h TTL) sat above the source-level cache with no invalidation signal from the callback-free auto-refresh path (`RefreshAsync`): a fleet rebuild re-advertised the stale RU set for up to an hour. New `IPrefixSourceService.ContentCommitted` event fires on every committed content change; `PrefixService` drops the projection when the default source changed and re-projects from the already-new cache (no duplicate push). 3 regression tests.

## #453 — server,protocol: OPEN in OpenConfirm is an FSM error (PR #459)
A malformed OPEN body in OpenConfirm escaped the handshake to RunAsync's parse-error catch and answered NOTIFICATION 2/x; RFC 4271 §8.2.2 makes any OPEN in OpenConfirm an FSM error regardless of body validity (same reasoning as #427 for Established). The handshake read now mirrors the #427 branch: exactly one NOTIFICATION 5/0, then Idle. Socket-pair red/green test.

## #454 — api: fetch budget for GET /api/as/{asn}/prefixes?count=true (PR #460)
The one cold-fetch GET endpoint #424 missed: a cold RIPEstat fetch (minutes-scale) pinned the request and its in-flight slot against the shutdown token only. Now bounded by the same linked `ExternalFetchBudget`; expiry answers a stable non-revealing 503 (a count cannot degrade to a partial list); shutdown still propagates. Hanging-provider test.

## #455 — api,server: create arms the shared-IP TCP-MD5 key via the #418 resolver (PR #461)
`HandleCreatePeer` armed the new row's key directly — last-writer-wins across the shared source IP, no disagreement warning — while delete/PATCH/bootstrap resolve via `RearmPeerIpMd5KeyAsync`/`ResolveSharedIpKey`. Create now goes through the same re-arm (deterministic ordinal pick). HTTP-driven regression test with a recording session manager.

## #456 — docs: D23 records the outbound AS_PATH re-origination (PR #462)
The D-catalog did not record that every advertised prefix is re-originated (AS_PATH = local ASN only, operator-stamped communities, no received-path propagation). D23 documents the decision and its RFC 4271 §9.1.2 consequence for receiving speakers' loop detection. Docs-only.

## #457 — protocol,server: oversized outbound UPDATE groups are dropped, not fatal (PR #463)
`SendUpdateBatchAsync` pre-validates the composed frame via `GetBufferSize` against the RFC 4271 §4.1 4096-byte maximum and drops an unsplittable group with an Error log — pre-fix the writer exception tore down the victim peer's session on every initial send. NLRI cannot overflow at the 100-per-UPDATE cap and attributes are constant per community group, so the #292-item-9 splitting reduces to this check. Initial-send red/green test + mirror-consistency assertions.

## Verification
- Per-PR: dotnet format / full solution build / full test suite; CI build-test + format-check + analyze (CodeQL) green; CodeQL merge-ref alerts 0 on every PR.
- Aggregate: 1052 tests green on the dev head; CodeRabbit review on THIS integration diff (per .coderabbit.yaml base-branch policy).

## Risk / changelog notes
- #454: a cold fetch slower than the 30s budget now answers 503 instead of holding the request for minutes.
- #455: on a source IP with differently-keyed siblings, the armed key is the deterministic ordinal pick (previously: last created won).
- #457: an unsplittable community group is skipped with an Error log (initial send previously reset the session).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TCP-MD5 key handling when multiple peers share an IP address.
  - ASN prefix-count requests now return a consistent 503 response when external data exceeds the configured time limit.
  - Corrected BGP error handling for unexpected OPEN messages.
  - Prevented oversized route updates from terminating active sessions; valid routes continue to be advertised.
  - Refreshed cached routing projections promptly when the default prefix source changes.

- **Documentation**
  - Documented outbound route re-origination, AS path handling, and community behavior.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -992,7 +992,12 @@ public sealed class ManagementApi : IHostedService, IDisposable
             normalizedIp, data.Asn, data.Description, asnLists, customPrefixes, data.CustomAsns ?? [], data.MaxPrefix,
             md5Password: string.IsNullOrEmpty(data.Md5Password) ? null : data.Md5Password);
         if (!string.IsNullOrEmpty(data.Md5Password))
-            _sessionManager.SetPeerMd5Key(normalizedIp, data.Md5Password);
+        {
+            // #455: resolve through the SAME shared-IP re-arm the delete/PATCH/bootstrap paths use —
+            // TCP keys by address (#36), so arming the new row's key directly silently overrode a
+            // sibling's key on the same IP (last-writer-wins, no disagreement warning).
+            await RearmPeerIpMd5KeyAsync(normalizedIp);
+        }
 
         _logger.LogInformation("Created peer {Ip} AS{Asn} ({Id}): {Subs} lists, {Prefixes} custom prefixes, {Asns} custom AS",
             normalizedIp, data.Asn, saved.Id, asnLists.Count, customPrefixes.Count, data.CustomAsns?.Count ?? 0);
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1522,10 +1522,24 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         if (countOnly)
         {
+            // #454: same wall-clock budget as /api/asn-lists (#424) — a cold RIPEstat fetch is
+            // minutes-scale and must not pin the request (and its in-flight slot) for the full
+            // timeout×retries chain. Unlike the two endpoints #424 covered, a count cannot
+            // degrade to a partial list, so budget expiry answers a stable 503; shutdown still
+            // propagates.
+            using var budget = CancellationTokenSource.CreateLinkedTokenSource(_shutdownCts.Token);
+            budget.CancelAfter(ExternalFetchBudget);
             try
             {
-                var count = await _prefixService.GetPrefixCountAsync(asn, _shutdownCts.Token);
+                var count = await _prefixService.GetPrefixCountAsync(asn, budget.Token);
                 return ApiResponse.Ok(new { asn, prefixCount = count });
+            }
+            catch (OperationCanceledException) when (!_shutdownCts.IsCancellationRequested)
+            {
+                _logger.LogWarning(
+                    "GET /api/as/{Asn}/prefixes exceeded the {Seconds:0}s external-fetch budget",
+                    asn, ExternalFetchBudget.TotalSeconds);
+                return ApiResponse.Error("The upstream prefix source did not respond in time", 503);
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
--- a/BGPLite.Providers/IPrefixSourceService.cs
+++ b/BGPLite.Providers/IPrefixSourceService.cs
@@ -48,4 +48,15 @@ public interface IPrefixSourceService
     /// interval selection in the auto-refresh timer.
     /// </summary>
     bool SourceSupportsConditional(string sourceName);
+
+    /// <summary>
+    /// #452: fired AFTER a source's freshly loaded content is committed to the cache and the load
+    /// detected an actual change — on EVERY change-detecting entry point, including
+    /// <see cref="RefreshAsync"/> (which deliberately never fires the <c>onSourceChanged</c>
+    /// BGP-push callback). Cache-layer consumers that project a source's list into their own
+    /// longer-TTL cache (the RU projection in <c>PrefixService</c>) subscribe to invalidate
+    /// their projection, so a fleet rebuild always re-projects from the committed content.
+    /// Fired off the per-source gate; handlers must be cheap and non-blocking.
+    /// </summary>
+    event Action<string>? ContentCommitted;
 }
--- a/BGPLite.Providers/PrefixService.cs
+++ b/BGPLite.Providers/PrefixService.cs
@@ -49,6 +49,28 @@ public sealed class PrefixService : IPrefixService
         // pipeline) so peer-controlled failures cannot open the breaker gating operator sources;
         // falls back to the operator provider when not wired (tests).
         _userSourceHttpProvider = userSourceHttpProvider ?? httpProvider;
+        // #452: a committed content change to the DEFAULT source must invalidate the RU projection
+        // immediately — the auto-refresh path (RefreshAsync) never fires the push callback, so
+        // without this the fleet rebuild triggered afterwards re-reads the stale fast path for up
+        // to the RU TTL. The handler is a Volatile.Write: cheap, non-blocking, safe off the
+        // per-source gate (LoadCachedAsync fires the event only after releasing it).
+        // Guarded: tests compose PrefixService WITHOUT a source service (the interface sits only
+        // on the RU paths), passing null — those compositions simply get no invalidation signal.
+        if (prefixSources is not null)
+            _prefixSources.ContentCommitted += OnSourceContentCommitted;
+    }
+
+    /// <summary>
+    /// #452 handler for <see cref="IPrefixSourceService.ContentCommitted"/>: drops the cached RU
+    /// projection when the changed source is the one it is projected from. The next
+    /// <see cref="GetRuPrefixesAsync"/> takes the gate path, re-projects from the already-new
+    /// source-level cache, and reports no change — so the fleet push is not duplicated.
+    /// </summary>
+    private void OnSourceContentCommitted(string sourceName)
+    {
+        if (!string.Equals(sourceName, _config.DefaultPrefixSource, StringComparison.Ordinal))
+            return;
+        Volatile.Write(ref _ruCache, null);
     }
 
     /// <summary>Default per-fetch budget for peer-supplied URL sources (#320).</summary>
--- a/BGPLite.Providers/PrefixSourceService.cs
+++ b/BGPLite.Providers/PrefixSourceService.cs
@@ -26,6 +26,10 @@ public sealed class PrefixSourceService : IPrefixSourceService
     // RefreshAsync (which would see already-new data and report unchanged) — leaving established peers
     // on stale routes. The callback is wired in Program.cs to ISessionManager.RefreshAllEstablishedAsync.
     private readonly Func<string, Task>? _onSourceChanged;
+    // #452: fired on EVERY committed content change (including RefreshAsync, which never fires
+    // _onSourceChanged). The RU projection's cache-layer consumer (PrefixService) subscribes to
+    // invalidate its projection so the next rebuild re-projects from the committed content.
+    public event Action<string>? ContentCommitted;
     // #85: pre-built name→source lookup (replaces per-call FirstOrDefault linear scan).
     private readonly Dictionary<string, PrefixSourceConfig> _sourcesByName;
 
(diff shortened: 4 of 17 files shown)


ruhex:
Per-finding verdicts on the CodeRabbit review of head `7a14a57` (the third finding — `AdvertisedPrefixCount` alignment — is **fixed** via #465, now included in this PR):

**1. “Make the shared-IP re-arm atomic with the peer save” (ManagementApi.cs) — won't-fix (pre-existing pattern; final state provably converges).**
The save-then-re-arm shape is the architecture #418 introduced and reviewed for the PATCH and bootstrap paths; this PR only routes create through the same seam. On the atomicity claim: two concurrent creates on one IP cannot both observe a partial view at the *final* re-arm — each re-arm runs after its own save, so “rearmA < saveB and rearmB < saveA” is a contradiction; at least the last-committed row's re-arm reads every row committed before it and arms the deterministic ordinal minimum. Only a *transient* window can carry a non-minimum key — and that window existed identically before this PR (create armed the new key unconditionally) and exists identically on the PATCH path. Making the re-arm transactional with the save would require holding a cross-request lock over the store write + the kernel key update — a redesign of #418's seam, out of scope for a batch of LOW findings. Distinct keys on one shared source IP are already a warned misconfiguration (the resolver logs it).

**2. “OPEN-confirm test harness: Socket.Send/Receive partial I/O” (OpenConfirmFsmErrorTests.cs) — won't-fix (consistent with the established harness).**
The socket-pair helpers deliberately mirror the existing ``MalformedUpdateResilienceTests`` harness (same Send/DrainAsync/ParseAll idiom) — hardening only the new file would leave the same property untested-and-unhardened in the older one. On the substance: every frame sent is ≤ 4096 bytes to a loopback socket with an empty send buffer and a blocking-mode socket, where a partial send is not an observable failure mode in practice, and the drain side tolerates arbitrary TCP segmentation (ParseAll loops over complete frames and stops at a partial tail, re-polling until the deadline). No flake has been observed across the suite's repeated CI runs. If the harness ever grows a real partial-write flake, the right fix is one shared hardened helper, not a divergent copy in this PR.

ruhex:
@coderabbitai review

PR #472, opened 2026-09-03
chore: integrate dev — MP-BGP error policy and cache gate fixes #466–#468

### LINKED ISSUE #466: protocol,server: MP IPv4/Unicast capability is advertised when the peer offers it, but MP_REACH/MP_UNREACH AFI=1 handling does not exist — negotiated IPv4 MP updates are silently discarded (RFC 4760 §8)
## Found by
External audit 2026-09-03 (MP-BGP deep dive). Verified against current `main`; sources untouched, full suite 1052/1052 green at filing time.

## Summary
`BgpSession.SendOpenAsync` answers a peer's MP IPv4/Unicast offer with our own advertisement:

```csharp
// Only advertise MP IPv4/Unicast if peer specifically supports IPv4/Unicast
if (PeerHasMpIpv4Unicast(remoteOpen.Capabilities))
    capabilities.Add(BgpCapabilityInfo.MultiprotocolIpv4Unicast());
```
(`BGPLite.Server/BgpSession.cs:1679-1681`; probe at `:1719-1730`.)

The inbound path has no AFI=1 handling at all: `BgpMessageReader.ParseUpdate` routes every MP_REACH_NLRI/MP_UNREACH_NLRI to the IPv6-only decoders (`BGPLite.Protocol/BgpMessageReader.cs:287-303`), and both reject `afi != 2` with `BgpParseException` (`BGPLite.Protocol/MpReachCodec.cs:70-73`, `:134-139`). That is a body error, so the whole UPDATE is discarded and the session stays up (D17 catch, `BgpSession.cs:765-803`) — the negotiated IPv4 MP routes are silently dropped.

Outbound is unaffected (IPv4 always rides classic NLRI in `SendRoutesAsync`), so the advertised capability is pure liability: we promise receive support we do not have.

## RFC requirement
- RFC 4760 §8: "each such speaker MUST advertise to the other (via the Capability Advertisement mechanism) the capability … to support that particular \<AFI, SAFI\> route" — bi-directional MP exchange presumes the advertisement is true. By advertising, BGPLite creates exactly the contract it then violates.
- RFC 5492 §3: "a given capability can be used on a peering if that capability has been advertised by both peers" — once we advertise, the peer is entitled to send IPv4 NLRI via MP_REACH.

## Consequences
- A conformant peer that prefers uniform MP carriage sends IPv4 NLRI via MP_REACH AFI=1/SAFI=1; BGPLite discards every such UPDATE — including any classic NLRI it may also carry (RFC 7606 §2 treat-as-withdraw is whole-UPDATE). Routes vanish with no NOTIFICATION and a healthy-looking session; only a Warning log line and the `UpdateRejected` counter hint at it.
- Same advertise-without-implementing class as #318 (Graceful Restart), resolved there by removing the advertisement (D6).

## Fix shape
Stop advertising MP IPv4/Unicast in `SendOpenAsync` until AFI=1 MP_REACH/MP_UNREACH is implemented (the D6 pattern). IPv4/Unicast needs no MP capability — it is the default family and BGPLite carries it via classic NLRI regardless. Implementing AFI=1 decode into the classic pipeline is the long-term fix; either way the advertisement must not precede the implementation.

## Severity
HIGH — silent route loss under a capability negotiation BGPLite itself completes; reachability requires a peer that both offers the capability and actually uses MP carriage for IPv4 (legal per RFC 5492/4760), so it is configuration-gated, not attack-driven.

## Regression test
Peer OPEN carries MP IPv4/Unicast (code 1, AFI=1/SAFI=1) → assert the session's OPEN does NOT echo that capability. Follow-up once AFI=1 lands: peer sends IPv4 NLRI via MP_REACH → routes install into the table.

### LINKED ISSUE #467: protocol,server: MP_REACH/MP_UNREACH bypass the mandatory-attribute policy — flags unvalidated, duplicates keep the session up (RFC 7606 §3(g)), unparseable MP leaves stale IPv6 routes (§3(j))
## Found by
External audit 2026-09-03 (MP-BGP deep dive); post-#287/#290 follow-up. The mandatory-attribute policy never sees MP_REACH/MP_UNREACH. Verified against current `main`.

## Summary
`BgpMessageReader.ParseUpdate` extracts MP attributes from the attribute stream before the generic list is built (`BGPLite.Protocol/BgpMessageReader.cs:287-303` — decoded and `continue`d, never added to `attributes`). Everything downstream of `ParseRouteAttributes` — flag/shape validation (`UpdateCodec.ValidateAttributeShape`), the §3(g) duplicate handling, the treat-as-withdraw machinery — therefore never applies to MP_REACH/MP_UNREACH. Three concrete defects:

1. **Flags unvalidated.** MP_REACH/MP_UNREACH are optional non-transitive (RFC 4760 §4/§5). An UPDATE carrying either with the Transitive bit set is malformed and is accepted today: `ParseAttribute` rejects only the reserved 0x08 bit (`BgpMessageReader.cs:341-346`), and no other check ever sees these attributes.
2. **Duplicates keep the session up.** The reader detects duplicates and throws `BgpParseException(UpdateMessageError)` (`BgpMessageReader.cs:289-294`, `:297-301`), which lands in the D17 body-error catch — message discarded, session stays up, no NOTIFICATION. The in-code comment at `UpdateCodec.cs:313-315` already acknowledges the prescribed response ("duplicate → NOTIFICATION, session reset") is not implemented; D17 cites §3(g) for the general discard rule but omits the MP carve-out.
3. **Unparseable MP value leaves stale routes.** A decode failure (bad AFI/SAFI, truncated MP NLRI, bad next-hop length) discards the UPDATE only. IPv6 routes the peer installed via an earlier good UPDATE survive indefinitely — the #288 stale-route hazard in its MP form: a peer can send a malformed MP attribute instead of a proper MP_UNREACH and keep stale routes active with the session looking healthy.

## RFC requirement (verified against the text)
- RFC 7606 §3(g): "If the MP_REACH_NLRI attribute or the MP_UNREACH_NLRI [RFC4760] attribute appears more than once in the UPDATE message then a NOTIFICATION message MUST be sent with the Error Subcode 'Malformed Attribute List'." → duplicates = NOTIFICATION 3/1 + session reset. RFC 7606 does not revise the MP attributes, so RFC 4271 §6.3 baseline applies to the rest.
- RFC 7606 §3(j): when the NLRI field and/or MP_REACH/MP_UNREACH cannot be successfully parsed, "the 'session reset' approach (or the 'AFI/SAFI disable' approach) MUST be followed". → the current discard-only behavior matches neither.

## Fix shape
Split MP-attribute error handling out of the D17 keep-alive catch:
- duplicate MP attribute → NOTIFICATION 3/1 (Malformed Attribute List) + teardown, per §3(g);
- flag conflicts / unparseable MP value → either session reset (RFC 4271 baseline) or the §3(j) "AFI/SAFI disable" approach: treat the peer's previously installed routes for that AFI/SAFI as withdrawn and stop accepting the family for the session. Either closes the stale-route hole; the chosen deviation must be recorded in DESIGN_DECISIONS (D17 needs the §3(g) MP carve-out noted either way);
- hardening (RFC 2545 §3): validate the IPv6 next hop — a global address is required; `::`, multicast, and a lone link-local are invalid. Today `DecodeMpReachV6` accepts any 16 bytes (`MpReachCodec.cs:85-87`) and the value lands in the route table unvalidated (outbound is unaffected — D22 always sends the configured next hop — so this is table-quality/robustness).

## Severity
HIGH — a peer can pin stale IPv6 routes past their withdrawal with one malformed attribute; duplicates and flag conflicts are protocol-conformance failures with session-state consequences.

## Regression test
- UPDATE with two MP_REACH_NLRI attributes → exactly one NOTIFICATION with Error=3, SubError=1 (Malformed Attribute List), session resets, owned routes flushed.
- MP_REACH with the Transitive flag set → rejected per the settled policy (not silently accepted).
- Install an IPv6 route via a valid MP_REACH → send an UPDATE whose MP_REACH is malformed (truncated NLRI) → the previously installed route is gone (withdrawn or session reset), behavior pinned per the settled decision.
- MP_REACH next hop `::` / multicast / link-local-only → rejected (Invalid Network Field).

### LINKED ISSUE #468: providers: a cancelled waiter can orphan the UserSourceCache gate mid-load — duplicate concurrent fetch of one URL and an older snapshot can overwrite a newer one
## Found by
External audit 2026-09-03 (cancellation/concurrency pass). Third member of the #358 (orphan-lock hygiene) / #450 (inflight invariant) family — this path predates neither and is not covered by either.

## Summary
In `UserSourceCache.GetOrLoadAsync` (`BGPLite.Providers/UserSourceCache.cs`), a caller cancelled **while queued** pair-removes the shared gate:

```csharp
catch (OperationCanceledException) when (ct.IsCancellationRequested)
{
    ...
    ((ICollection<KeyValuePair<string, SemaphoreSlim>>)_locks).Remove(new KeyValuePair<string, SemaphoreSlim>(url, gate));  // :122
    throw;
}
```

The removal fires while the FIRST loader may still be running (it holds the removed gate instance) or while other waiters are queued on it. A subsequent caller `GetOrAdd`s a fresh gate (`:100`) and starts a second concurrent `loadAsync` for the same URL — exactly what the `_inflight`/#450 invariant ("the sweeps never remove a gate whose loaders are still inside or queued", `:57-59`, `:283-285`) exists to prevent. The cancel path bypasses that invariant.

If the newer load finishes first, `_cache[url]` is written unconditionally by whichever loader lands last (`:163`, fresh timestamp either way) — the older-started loader can overwrite the newer snapshot, so sessions refreshing the same URL around the window can observe different prefix lists.

The comment at `:118-121` acknowledges the duplicate-fetch tradeoff for the case where the cancelled caller was the ONLY participant; the defect is that it also fires when it is not, and that the final write has no ordering guard.

## Current behavior
Cancelled queued waiter → gate instance removed under an active loader → second gate minted → concurrent duplicate fetch of one URL; last-writer-wins on the cache entry.

## Fix shape
Make the cancel path respect the same invariant the sweeps follow: decrement `_inflight` first, then pair-remove the gate ONLY when the caller was the last participant (`_inflight` reached 0); otherwise leave the gate in place for the loader/waiters and let the sweeps reclaim the (bounded) orphan later — which requires the sweeps to also collect gates with no cache entry and no active loaders, so #358's leak does not return. Optionally guard the cache write with a monotonic attempt marker so a late older load cannot clobber a newer snapshot.

## Severity
MEDIUM — needs a cancellation timed mid-queue; consequences are a duplicate upstream fetch and a transient snapshot regression between sessions, not data corruption (either list is well-formed for the same URL).

## Regression test
Barrier test: loader A holds the gate (fetch blocked on a signal); waiter B queues and is cancelled; caller C arrives → assert C waits on A's gate (no second load starts while A is active), then release A and assert exactly one cache write occurred (the completed snapshot, not an interleaved overwrite).

ruhex:
Closes #466
Closes #467
Closes #468

## #466 — stop advertising the MP IPv4/Unicast capability (PR #469)
The OPEN echoed a peer's MP IPv4/Unicast (AFI=1/SAFI=1) offer while the inbound path has no MP_REACH/MP_UNREACH AFI=1 handling at all: RFC 5492 §3 / RFC 4760 §8 made the capability usable, and every negotiated IPv4 MP UPDATE was discarded whole (treat-as-withdraw) with the session looking healthy — silent route loss. Fix: never advertise MP IPv4/Unicast (D24, the D6 pattern); IPv4/Unicast rides classic NLRI. Verified: red/green capability test, full suite, CodeQL clean.

## #467 — enforce the RFC error policy for MP_REACH/MP_UNREACH (PR #470)
MP attributes were extracted before the mandatory-attribute pipeline, so RFC 7606's carve-out was bypassed: flags conflicts accepted, duplicates silently discarded, unparseable MP values left stale peer routes in the table. Fix: duplicate MP attribute → NOTIFICATION 3/1 + session reset (RFC 7606 §3(g) MUST); flags conflict → 3/4 + reset (RFC 4271 §6.3 baseline); unparseable MP value → the §3(j) "AFI/SAFI disable" choice (withdraw the peer's accepted IPv6 routes, ignore the family for the session, keep the session up — recorded in D17); MP_REACH next hop must be a global IPv6 address (RFC 2545 §3), otherwise route-level exclusion. Verified: 4/5 new session tests proven red pre-fix, full suite 1065/1065, CodeQL clean.

## #468 — a cancelled waiter must not split the UserSourceCache gate (PR #471)
A waiter cancelled while queued removed the shared per-URL gate out from under the still-running first loader; the next caller minted a second gate and fetched the same URL concurrently, and the older load could overwrite the newer snapshot (different sessions observing different prefix lists). Fix: register in _inflight before taking the gate; the gate is removed by the LAST leaver only — no mid-load removal, #358 orphan hygiene preserved. Verified: red/green concurrency test, both #358 tests unchanged and green, full suite 1066/1066, CodeQL clean.

## Summary by CodeRabbit

- **New Features**
  - Added MP_REACH_NLRI and MP_UNREACH_NLRI support for IPv4 and IPv6 unicast routing.
  - IPv4 multiprotocol capability is now advertised automatically.
  - Added address-family-specific route removal and IPv6 next-hop validation.

- **Bug Fixes**
  - Improved malformed multiprotocol attribute handling, including session resets and IPv6 family disablement.
  - IPv4 updates continue processing when IPv6 multiprotocol data is disabled.
  - Prevented duplicate concurrent URL loads after cancellation.

- **Documentation**
  - Documented multiprotocol error handling and IPv4 capability behavior.

### DIFF
--- a/BGPLite.Protocol/BgpMessageReader.cs
+++ b/BGPLite.Protocol/BgpMessageReader.cs
@@ -267,6 +267,10 @@ public static class BgpMessageReader
         var attributes = new List<PathAttribute>();
         MpReachCodec.MpReachV6? mpReachV6 = null;
         IReadOnlyList<IpPrefix>? mpUnreachV6 = null;
+        MpReachCodec.MpReachV4? mpReachV4 = null;
+        IReadOnlyList<IpPrefix>? mpUnreachV4 = null;
+        var mpReachSeen = false;
+        var mpUnreachSeen = false;
         if (attrsLen > 0)
         {
             var attrsEnd = offset + attrsLen;
@@ -303,20 +307,28 @@ public static class BgpMessageReader
                             BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.AttributeFlagsError,
                             sessionResetRequired: true);
 
+                    // #466: BOTH supported families decode here — IPv4/Unicast (AFI=1/SAFI=1) into
+                    // the classic NLRI pipeline and IPv6/Unicast (AFI=2/SAFI=1) into MP_REACH_NLRI.
+                    // §3(g) duplicate detection is per ATTRIBUTE TYPE: a second instance of the
+                    // type is a session reset regardless of which family it names.
                     if (attr.TypeCode == MpReachCodec.MpReachNlriType)
                     {
                         // RFC 7606 §3(g): "If the MP_REACH_NLRI attribute or the MP_UNREACH_NLRI
                         // [RFC4760] attribute appears more than once in the UPDATE message then a
                         // NOTIFICATION message MUST be sent with the Error Subcode 'Malformed
                         // Attribute List'." — a session reset, NOT the D17 keep-alive discard.
-                        if (mpReachV6 is not null)
+                        if (mpReachSeen)
                             throw new BgpParseException("Duplicate MP_REACH_NLRI attribute (RFC 7606 §3(g))",
                                 BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAttributeList,
                                 sessionResetRequired: true);
-                        ScopeMpValueOrThrow(attr.Data);
+                        mpReachSeen = true;
+                        var reachIsV4 = ScopeMpValueOrThrow(attr.Data);
                         try
                         {
-                            mpReachV6 = MpReachCodec.DecodeMpReachV6(attr.Data);
+                            if (reachIsV4)
+                                mpReachV4 = MpReachCodec.DecodeMpReachV4(attr.Data);
+                            else
+                                mpReachV6 = MpReachCodec.DecodeMpReachV6(attr.Data);
                         }
                         catch (BgpParseException ex)
                         {
@@ -327,14 +339,18 @@ public static class BgpMessageReader
                         continue;
                     }
 
-                    if (mpUnreachV6 is not null)
+                    if (mpUnreachSeen)
                         throw new BgpParseException("Duplicate MP_UNREACH_NLRI attribute (RFC 7606 §3(g))",
                             BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAttributeList,
                             sessionResetRequired: true);
-                    ScopeMpValueOrThrow(attr.Data);
+                    mpUnreachSeen = true;
+                    var unreachIsV4 = ScopeMpValueOrThrow(attr.Data);
                     try
                     {
-                        mpUnreachV6 = MpReachCodec.DecodeMpUnreachV6(attr.Data);
+                        if (unreachIsV4)
+                            mpUnreachV4 = MpReachCodec.DecodeMpUnreachV4(attr.Data);
+                        else
+                            mpUnreachV6 = MpReachCodec.DecodeMpUnreachV6(attr.Data);
                     }
                     catch (BgpParseException ex)
                     {
@@ -360,7 +376,9 @@ public static class BgpMessageReader
             PathAttributes = attributes,
             Nlri = nlri,
             MpReachV6 = mpReachV6,
-            MpUnreachV6 = mpUnreachV6
+            MpUnreachV6 = mpUnreachV6,
+            MpReachV4 = mpReachV4,
+            MpUnreachV4 = mpUnreachV4
         };
     }
 
@@ -368,12 +386,12 @@ public static class BgpMessageReader
     /// #472 review: scope MP value failures to the offending AFI/SAFI tuple. A value too short
     /// to name its tuple cannot be scoped to any family — the RFC 7606 §3(j) fallback is the
     /// session reset. A tuple this speaker does not support was never negotiated (RFC 4760 §8):
-    /// its arrival is not a parse failure of the supported family, so it is answered with a
-    /// plain keep-alive body error — the whole UPDATE is discarded (D17) and the supported
-    /// family stays enabled. Only a supported tuple's decode failure reaches the decoder, whose
-    /// <see cref="BgpMpParseException"/> marks the AFI/SAFI disable.
+    /// its arrival is not a parse failure of a supported family, so it is answered with a plain
+    /// keep-alive body error — the whole UPDATE is discarded (D17) and the supported families
+    /// stay enabled. Supported tuples (#466): AFI=1/SAFI=1 and AFI=2/SAFI=1.
     /// </summary>
-    private static void ScopeMpValueOrThrow(ReadOnlySpan<byte> value)
+    /// <returns>true for the IPv4/Unicast tuple, false for IPv6/Unicast.</returns>
+    private static bool ScopeMpValueOrThrow(ReadOnlySpan<byte> value)
     {
         if (value.Length < 3)
             throw new BgpParseException(
@@ -383,10 +401,14 @@ public static class BgpMessageReader
 
         var afi = BinaryPrimitives.ReadUInt16BigEndian(value);
         var safi = value[2];
-        if (afi != MpReachCodec.AfiIpv6 || safi != MpReachCodec.SafiUnicast)
-            throw new BgpParseException(
-                $"MP_REACH/MP_UNREACH for an unsupported address family: AFI={afi}, SAFI={safi}",
-                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.InvalidNetworkField);
+        if (afi == MpReachCodec.AfiIpv4 && safi == MpReachCodec.SafiUnicast)
+            return true;
+        if (afi == MpReachCodec.AfiIpv6 && safi == MpReachCodec.SafiUnicast)
+            return false;
+
+        throw new BgpParseException(
+            $"MP_REACH/MP_UNREACH for an unsupported address family: AFI={afi}, SAFI={safi}",
+            BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.InvalidNetworkField);
     }
 
     private static (PathAttribute attr, int consumed) ParseAttribute(ReadOnlySpan<byte> data)
--- a/BGPLite.Protocol/BgpMessageReader.cs
+++ b/BGPLite.Protocol/BgpMessageReader.cs
@@ -334,7 +334,7 @@ public static class BgpMessageReader
                         {
                             // RFC 7606 §3(j): a SUPPORTED tuple whose value cannot be decoded takes
                             // the "AFI/SAFI disable" choice, scoped to that family (D17).
-                            throw new BgpMpParseException(ex.Message, ex.SubErrorCode, ex);
+                            throw new BgpMpParseException(ex.Message, reachIsV4, ex.SubErrorCode, ex);
                         }
                         continue;
                     }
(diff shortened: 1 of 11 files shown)


ruhex:
Per-finding verdicts on the review of 90d4a9a — all three accepted findings are fixed by #473 (now the PR head, ec5069c); two findings answered won't-fix with rationale:

1. **Flag mask / Partial bit — FIXED (ec5069c).** The conflict mask now includes `Partial`; valid flag bytes are exactly `0x80`/`0x90`. Covered by `PartialMpReachFlags_TearDownSession_WithNotification_3_4`.
2. **Session reset for parseable MP values with flag conflicts — WON'T-FIX.** RFC 7606 §3 explicitly leaves MP_REACH_NLRI/MP_UNREACH_NLRI outside its revisions ("The MP_REACH_NLRI and MP_UNREACH_NLRI attributes [RFC4760] … the procedures described in this section do not apply" — §3's scope), so the suggested treat-as-withdraw for their flag conflicts would apply a revision that excludes them. The applicable text is the RFC 4271 §6.3 baseline: recognized attribute with conflicting flags → Attribute Flags Error (3/4) + session reset. Only §3(g) (duplicates → 3/1 MUST) and §3(j) (unparseable → reset or disable) carve MP attributes out of the baseline, and both are implemented as such.
3. **Tuple-scoped recovery — FIXED (ec5069c).** `ScopeMpValueOrThrow` now separates three cases: unsupported tuple (never negotiated, RFC 4760 §8) → keep-alive whole-UPDATE discard, session and supported family untouched; supported tuple with an undecodable value → BgpMpParseException → AFI/SAFI disable; a value too short to name its tuple → §3(j) fallback session reset. Covered by `UnsupportedMpFamily_DiscardsTheUpdateOnly_SessionAndFamilyStayUp` and `UnreadableMpTuple_ResetsSession_CannotBeScopedToAFamily`.
4. **Link-local half of 32-byte next hops — WON'T-FIX.** The second 16 bytes are deliberately unconsumed (the codec reads only the global half), and the consumed half IS validated (`IsGlobalUnicastNextHop`, RFC 2545 §3). Wiring the ignored, advisory field into validation would escalate harmless garbage in a field we never use into an UPDATE discard or a family disable — route loss with no operational benefit. Recorded as an accepted limitation here rather than silently skipped.
5. **`_installedPrefixes` leak after disable — FIXED (ec5069c).** `DisablePeerMpV6` drops the family's keys; `RemoveAllOwnedBy` intentionally does not raise `EntryOwnershipLost` for self-discarded keys (that event means "a REPLACING owner took it"). Covered by `AfterFamilyDisable_ThePrefixCapBudgetIsReturned`.
6. **"Atomic gate lifecycle" for #468 — WON'T-FIX (already guaranteed).** The register-before-take ordering makes a lock unnecessary: a participant increments `_inflight` BEFORE `_locks.GetOrAdd`, and a gate is removed only when the count reaches 0 — so between any participant's registration and its acquisition the count is ≥1 and no removal can run; two gates for one URL can never coexist with active loaders. A shared lifecycle record would add synchronization the ordering already proves unnecessary.

ruhex:
@coderabbitai review

Issue #254, opened 2026-08-26
server: RefreshRoutesAsync bounding token is CancellationToken.None — the #160 bound does not work; no debounce of stacked refreshes


ruhex:
## What is wrong

`RefreshRoutesAsync(CancellationToken ct = default)` (`BgpSession.cs:89`) — `default` is `CancellationToken.None`, **not** the session's `_cts.Token` claimed by the comment at `:98-100`. All management-API callers pass no token: `BgpServer.cs:349` (`RefreshPeerAsync`), `BgpServer.cs:385` (`RefreshAllEstablishedAsync`).

## Impact

- The #160 promise ("bounds how long a management-API caller blocks here") does not exist: while the initial send holds `_advertisedPrefixesLock` (RIPE fetch + thousands of batches), every `POST/PUT /api/peers*` mutation (`ManagementApi.cs:665,753,835,853,877`) and every `onSourceChanged` wave (`Program.cs:112-117`) enqueues a **non-cancellable** task that then runs a full withdraw+re-announce.
- N triggers in a row → the peer receives N sequential full dumps.
- Tasks are not cancelled even when the session tears down (they end only on a send error).

## Fix

- Normalize `ct == default → _cts.Token` at the method entry (or change signatures and pass tokens explicitly).
- Debounce/coalesce concurrent refresh requests per session (e.g. a pending-refresh flag or single-slot queue; later requests await the in-flight one instead of queueing).

## Acceptance

- Session teardown cancels a queued/running refresh.
- N rapid refresh triggers produce at most one (or one in-flight + one pending) wire dump, not N.

Refs: audit 2026-08-26; #160.

PR #500, opened 2026-09-03
fix(protocol): reject a zero-length AS_PATH at the eBGP policy layer (#486)

### LINKED ISSUE #486: protocol: a zero-length AS_PATH is accepted as valid inbound — settle the policy
## Found by
External audit 2026-09-04 (attribute pass). Not a #238 duplicate: #238 settled the empty SEGMENT case (`segmentLength == 0` inside a non-empty attribute → rejected as Malformed AS_PATH); this is the zero-length ATTRIBUTE — `data.Length == 0`, the segment loop never entered, no rejection anywhere.

## Summary
`AttributeHelper.ReadAsPath` returns `[]` for a zero-length AS_PATH without error (`ReadPathData` with `data.Length == 0` → the `while (offset + 2 <= data.Length)` loop body never runs → empty array), `asPathSeen` is set (`UpdateCodec.cs:360-362`), the mandatory-attribute check passes, and the routes install with an empty path.

RFC 4271 §5.1.2 describes AS_PATH segments as carrying "one or more AS numbers"; on an eBGP session the sender's ASN must be present, so a zero-length AS_PATH is malformed in practice (major implementations reject it as Malformed AS_PATH, subcode 11). RFC requirement not established verbatim — "empty on eBGP = malformed" is an interpretation of §5.1.2 plus implementation practice (BGPLite is eBGP-only; an empty path is legal only for iBGP, which BGPLite does not have) — hence a settle-the-policy issue rather than a straight bug.

Impact is bounded by D23 (peer-learned routes are never re-advertised; the AS-loop exclusion at `BgpSession.cs:1126` simply never matches an empty path) — this is table quality / robustness, not a leak.

## Fix shape (decision required)
Either reject the zero-length AS_PATH as Malformed AS_PATH (treat-as-withdraw) in the inbound path, or record the acceptance as a D-entry with the rationale.

## Severity
LOW.

## Regression test
UPDATE with a zero-length AS_PATH attribute + NLRI → behavior pinned per the settled decision (rejected with 3/11 treat-as-withdraw, or installed and documented).

ruhex:
Closes #486   # linkage only — the integration PR aggregates the closing keywords

## Problem
`UpdateCodec.ParseRouteAttributes` accepted a zero-length AS_PATH: `ReadPathData`'s segment loop never ran, `asPathSeen` was set, and routes installed with an empty path — which the AS-loop exclusion (RFC 4271 §9.1.2) can never match (D23), and which no conformant eBGP sender produces.

## Root Cause
#238 rejected the empty-SEGMENT case; the empty-ATTRIBUTE case (the segment loop's zero-iteration path) was never covered.

## Fix
Policy-layer rejection in `ParseRouteAttributes` (`3/11` Malformed AS Path → the caller's treat-as-withdraw). The CODEC stays encoding-neutral: a zero-length attribute is the legal on-wire form of an empty path (RFC 4271 §5.1.2, iBGP), the writer's empty-path roundtrip (#248 review: "the writer must not emit what the reader rejects") stands, and AS4_PATH stays lenient (the RFC 6793 merge ignores an empty one). Decision recorded as **D25** in DESIGN_DECISIONS.md.

## RFC
- RFC 4271 §5.1.2 (path segments carry "one or more AS numbers"; empty path only toward internal peers), §9.1.2 (loop detection over the path), §6.3 (Malformed AS_PATH). The verbatim eBGP prohibition is interpretation + implementation practice — recorded as a decision (D25), exactly what #486 asked ("settle the policy").

## Regression Test
`BgpMessageTests.AsPath_ZeroLengthAttribute_IsMalformedOnEbgp_PolicyLayer` — asserts `ParseRouteAttributes` throws 3/11 for a zero-length AS_PATH and, in the same test, that the codec layer still decodes it (policy/codec split pinned). **Red/green proven.**

## Verification
- `dotnet format` clean; `dotnet build BGPLite.sln` 0 warnings/0 errors
- `dotnet test BGPLite.sln`: 1087/1087 (baseline 1086 + the new test)

## Behavior Change
An UPDATE carrying a zero-length AS_PATH is now treated-as-withdraw (session stays up, UpdateRejected counted) instead of installing routes with an empty path.

## Deviation from Issue Proposal
The issue suggested rejecting inside `ReadAsPath`; analysis showed that would break the writer/reader symmetry pinned by `AsPath_Empty_RoundtripsAsZeroLengthAttribute` (#248) — the rejection belongs at the eBGP policy layer, not the codec.

### DIFF
--- a/BGPLite.Protocol/UpdateCodec.cs
+++ b/BGPLite.Protocol/UpdateCodec.cs
@@ -358,6 +358,16 @@ public static class UpdateCodec
                         originSeen = true;
                         break;
                     case BgpConstants.Attribute.AsPath:
+                        // #486 (D25): a zero-length AS_PATH is a legal ENCODING (RFC 4271 §5.1.2
+                        // lets an originator send an empty path toward INTERNAL peers) but never a
+                        // valid eBGP path — BGPLite serves eBGP only, where the sender's own ASN
+                        // must be present. Rejected at the policy layer as Malformed AS_PATH
+                        // (treat-as-withdraw); the codec stays encoding-neutral so the writer's
+                        // empty-path roundtrip (#248 review) stands.
+                        if (attr.Data.Length == 0)
+                            throw new BgpNotificationException(
+                                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAsPath,
+                                "Invalid AS_PATH: an empty path is only valid toward internal peers (RFC 4271 §5.1.2)");
                         asPath = AttributeHelper.ReadAsPath(attr, fourByteAsnSession);
                         asPathSeen = true;
                         break;
--- a/BGPLite.Tests/BgpMessageTests.cs
+++ b/BGPLite.Tests/BgpMessageTests.cs
@@ -654,6 +654,33 @@ public class BgpMessageTests
     }
 
     [Fact]
+    public void AsPath_ZeroLengthAttribute_IsMalformedOnEbgp_PolicyLayer()
+    {
+        // #486 (D25): a zero-length AS_PATH is the on-wire form of an EMPTY path — legal only
+        // toward internal peers (RFC 4271 §5.1.2). BGPLite is eBGP-only, so the inbound policy
+        // layer rejects it as Malformed AS_PATH (treat-as-withdraw); the codec itself stays
+        // encoding-neutral (see AsPath_Empty_RoundtripsAsZeroLengthAttribute).
+        var update = new BgpUpdateMessage
+        {
+            PathAttributes =
+            [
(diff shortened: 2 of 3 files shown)


ruhex:
CI note: the first build-test run failed on `RefreshDebounceTests.ConcurrentRefreshes_CoalesceIntoOneCyclePlusOneLap` (Assert.InRange 1–2, actual 3) — a #254 debounce timing test with no interaction with this diff (protocol-layer only); it passed locally (1087/1087) and on the rerun with zero changes. Recorded as a CI-flakiness observation for the batch report (not fixed here — one issue : one PR).

Issue #489, opened 2026-09-03
chore: external audit 2026-09-04 — tracking issue (0 critical / 1 high / 5 medium / 7 low)


ruhex:
## Scope
External adversarial audit 2026-09-04 (follow-up to the 2026-09-03 audit tracked in #432): full pass over Protocol / Server / Routing / Providers / Api / Configuration, RFC conformance (4271/5492/6793/7606/7607/4760/2545/8092/1997/2918/4724/4486), concurrency/lifecycle, routing correctness, resource exhaustion. Core files read end-to-end; subsystem passes run by four independent sub-audits; load-bearing findings re-verified against the code (plus targeted .NET 10 runtime probes for CTS/semaphore semantics). Full suite 1074/1074 green at audit time.

## Audit verdict on the last 5 issues (#456–#468)
- **#466** (MP IPv4 advertise-without-receive) — **ALREADY FIXED correctly**: both halves implemented (receive + unconditional advertisement, D24); tests present.
- **#467** (MP attributes bypass the mandatory-attribute policy) — **ALREADY FIXED**: §3(g) duplicates → 3/1 + reset, flags → 3/4 + reset, §3(j) scoped per family (v6 disable / v4 reset), RFC 2545 §3 next-hop exclusion; D17 carve-out recorded. One adjacent residual filed as #484.
- **#468** (cancelled waiter splits the gate) — **PARTIALLY FIXED**: registration ordering is correct, but a TOCTOU remains in `ExitInflight`/sweeps → filed as #478.
- **#457** (>4096-byte outbound UPDATE) — **ALREADY FIXED** via pre-validation + drop-group + wire-truth count; regression tests present.
- **#456** (AS_PATH re-origination docs) — **ALREADY FIXED** (D23).

## New findings (0 critical / 1 high / 5 medium / 7 low)

| Issue | Sev | Area | Summary |
|---|---|---|---|
| #476 | HIGH | server,protocol | `MergeDuplicatePrefixes` drops `IsIpv4` — a duplicated IPv6 prefix tears down the initial send (flap loop) or announces a bogus IPv4 NLRI (`0.0.0.0/N`) |
| #477 | MEDIUM | server,config | `ConfigCommunityResolver`/`RouteAssembler` NRE on the documented-valid `PrefixSources:` null config |
| #478 | MEDIUM | providers | `ExitInflight` removes the gate unconditionally — TOCTOU residual of the #468 single-fetch invariant |
[...]

Issue #163, opened 2026-07-05
[bug/providers] RIPEstat provider has no stale-on-failure — transient outage drops all routes


ruhex:
## Bug

`PrefixService.GetPrefixesAsync` (RIPEstat) and `GetRuPrefixesAsync` have **no stale-on-failure fallback**:

```csharp
// BGPLite.Providers/PrefixService.cs:49-59 (GetPrefixesAsync) and :124-133 (GetRuPrefixesAsync)
// On a refresh after TTL expiry, if RIPEstat throws (429/5xx/network), the exception propagates —
// there is no try/catch that serves the expired-but-cached entry.
```

This is asymmetric with `PrefixSourceService` and `UserSourceCache`, which **both** explicitly retain and serve the last-good copy on refresh failure.

## Impact

- A 30-second RIPEstat hiccup causes **mass route withdrawal** for every peer the instant each ASN's TTL elapses — the most rate-limit-prone source is the *least* resilient.
- File/HTTP sources stay up during the same outage, so behavior is inconsistent and surprising to operators.
- Withdrawn prefixes propagate to peers as actual BGP withdrawals — real network impact, not just stale data.

## Fix

- Mirror the stale-on-failure pattern from `PrefixSourceService`: on refresh throw, serve the expired entry (with an optional `stale` indicator) and schedule a background re-fetch.
- Optionally cap stale-serving duration (e.g. serve stale up to 2× TTL, then give up).

## Scope

M. Provider-layer; covered by a test that injects a transient RIPEstat failure after TTL expiry and asserts the previous entry is served.

Refs: companion issue (cache stampede) — the two together make RIPEstat resilience coherent.

PR #503, opened 2026-09-04
chore: integrate dev — 2026-09-04 audit fixes #476–#488

### LINKED ISSUE #476: server,protocol: MergeDuplicatePrefixes drops IsIpv4 — a duplicated IPv6 prefix tears down the initial send or announces a bogus IPv4 NLRI
## Found by
External audit 2026-09-04 (adversarial routing pass). Verified against current `main`; full suite 1074/1074 green at filing time.

## Summary
`BgpSession.MergeDuplicatePrefixes` constructs the merged `Route` WITHOUT `IsIpv4` (`BGPLite.Server/BgpSession.cs:1558-1566`), and `Route.IsIpv4` defaults to `true` (`BGPLite.Routing/Route.cs:16`):

```csharp
merged[key] = new Route
{
    Prefix = existing.Prefix,
    PrefixLength = existing.PrefixLength,
    NextHop = existing.NextHop,
    AsPath = existing.AsPath,
    Communities = comms,
    LargeCommunities = large
};
```

The merge KEY carries the family (`:1551` — `(route.Prefix, route.PrefixLength, route.IsIpv4)`), so two IPv6 duplicates land exactly in this branch — and come out as `IsIpv4 = true`.

Duplicates are guaranteed reachable: `ExactUnionPrefixAggregator` deliberately keeps prefixes with different community sets in separate groups (`ExactUnionPrefixAggregator.cs:60-64`), so the same IPv6 prefix from two sources (e.g. the RU default + an ASN subscription — RIPE returns both families — or a custom v6 prefix) reaches `MergeDuplicatePrefixes` as a cross-group duplicate.

## Current behavior
The family split (`BgpSession.cs:1436-1441`) routes the corrupted route into `v4Routes`; `new IpPrefix(r.Prefix, r.PrefixLength, r.IsIpv4)` at `:1598` then:

- **length 33..128** → `ArgumentOutOfRangeException` from the ctor (`IpPrefix.cs:34-35`):
  - initial send: the AOORE reaches `RunAsync`'s generic catch (`:557`) → best-effort Cease → teardown → the peer reconnects into the same state → **permanent session flap**;
  - refresh path: caught at `RefreshCycleAsync:200-203`, but `WithdrawAllAsync` has already run → **the peer keeps zero routes** until a refresh that no longer throws;
- **length 0..32** (e.g. `2001:db8::/32`) → no throw: the ctor masks the low 32 bits (`IpPrefix.cs:39`) → **`0.0.0.0/32` (or `0.0.0.0/N`) goes on the wire as classic IPv4 NLRI** — blackhole injection into the peer's table.

IPv4 duplicates are unaffected (the default matches the family), which is why v4-only deployments never see it.

## Why existing protection is insufficient
No earlier dedup exists — the aggregator *produces* the cross-community duplicates; the only merge point is the buggy one. No test covers v6 duplicates (`MergeDuplicatePrefixes` is exercised for v4 only).

## Fix shape
One line: `IsIpv4 = existing.IsIpv4,` in the merged `Route` initializer.

## Severity
HIGH — wire corruption / session flap from a legitimate configuration (no attacker needed); fix is one line.

## Regression test
Send two routes with the same IPv6 prefix and different community sets through `SendRoutesAsync` → exactly one MP_REACH UPDATE with the union of the communities reaches the peer, the session stays Established, and no classic IPv4 NLRI for the prefix is emitted.

### LINKED ISSUE #477: server,config: ConfigCommunityResolver/RouteAssembler throw NRE on the documented-valid PrefixSources: null config
## Found by
External audit 2026-09-04 (composition pass). Verified against current `main` and by executing a probe against `BGPLite.Configuration` on .NET 10. Not a #327 duplicate: #327 added validation of malformed `PrefixSources` VALUES; this is the YAML-null case, which `Validate` still accepts.

## Summary
`AppConfig.Validate` tolerates a null `PrefixSources` (`PrefixSources ?? []` in its own checks, `AppConfig.cs:161`-ish) and `ConfigValidationTests.cs:253-259` explicitly blesses writing `PrefixSources:` (YAML-null) as "no sources" — "every runtime consumer treats it that way". Two consumers do not:

- `BGPLite.Server/ConfigCommunityResolver.cs:52` — `config.PrefixSources.ToDictionary(...)` without `?? []` (contrast `config.RipeStat?.AsnLists ?? []` two lines above at `:49`) → **ArgumentNullException at DI resolution → startup crash** with a stack trace that names neither the config key nor the fix;
- `BGPLite.Server/RouteAssembler.cs:156,158` — `prefixSources.Any(...)` on the same null → NRE on the outbound build path.

`ConfigLoader` does not normalize the field, so the null reaches both consumers verbatim (probe confirmed: `PrefixSources == null` after a `PrefixSources:` YAML document, and `ToDictionary` → `ArgumentNullException`).

## Current behavior
A config that passes `Validate()` and has a dedicated passing test crashes the process at startup (or NREs the outbound build) with a bare NRE/ANE.

## Fix shape
Preferred: normalize the whole `AppConfig` once right after `Validate()` (closes the class — the null-vs-[] asymmetry is one future consumer away from recurring). Minimal: `config.PrefixSources ?? []` at both consumers.

## Severity
MEDIUM — availability (startup crash on a documented-valid config) with a confusing diagnostic; not a correctness issue.

## Regression test
Compose `ConfigCommunityResolver`/`RouteAssembler` from an `AppConfig` with `PrefixSources = null` → construction succeeds, community resolution returns defaults, the outbound build contributes no prefix-source routes.

### LINKED ISSUE #478: providers: ExitInflight removes the gate unconditionally — a TOCTOU residual of the #468 single-fetch invariant
## Found by
External audit 2026-09-04 (cancellation/concurrency pass). Follow-up to #468 — the same invariant, a much narrower window. Found twice independently in the same audit (main auditor + providers sub-audit).

## Summary
The #468 fix registers loaders in `_inflight` before taking the gate and moves gate removal into `ExitInflight` — but `ExitInflight` pair-removes the gate UNCONDITIONALLY (`BGPLite.Providers/UserSourceCache.cs:180-186`):

```csharp
var remaining = _inflight.AddOrUpdate(url, 0, (_, c) => c - 1);
if (remaining > 0) return;
_inflight.TryRemove(new KeyValuePair<string, int>(url, 0));                       // can FAIL…
((ICollection<KeyValuePair<string, SemaphoreSlim>>)_locks).Remove(new(url, gate)); // …and this still runs
```

Interleaving: A decrements to 0 → C registers (+1) and takes gate G1 from `_locks` → A's `TryRemove(url, 0)` fails (the value is now 1) → A's `_locks.Remove((url, G1))` **succeeds anyway, removing the gate C now holds** → D arrives, `GetOrAdd` mints G2 → C and D are inside concurrently → duplicate concurrent `loadAsync` of one URL; both then write `_cache[url]` unconditionally (`:148`/`:155`) → an older response can overwrite a newer one for a full positive TTL.

The sweep paths have the identical check-then-act window (`_inflight.TryGetValue(key, active > 0) → skip; else remove gate` at `:223-229` and `:284-291`).

The doc comments at `:100-105`/`:169-178` claim "ExitInflight only ever removes a gate that nobody holds or queues on" — the code does not currently provide that guarantee.

## Current behavior
Nanosecond-window TOCTOU: one duplicate idempotent upstream GET + a transient snapshot flip between sessions reading the same URL — the exact #468 consequence class, rarer trigger.

## Fix shape
Remove the gate only when the `_inflight.TryRemove(KVP(url, 0))` succeeded (that CAS *is* "nobody else registered"); or stop removing gates in the exit path entirely — the entry cap already bounds the number of live gates — and let the sweeps (with the same guard) reclaim them.

Note: the barrier regression test proposed in #468 ("waiter B cancelled; caller C arrives → C waits on A's gate, no second load starts; exactly one cache write") was never added — the merged tests cover orphan-gate hygiene only. Add it with this fix.

## Severity
MEDIUM — needs a registration landing in a nanosecond window; consequences are a duplicate fetch and a transient snapshot regression, not data corruption.

## Regression test
Deterministic barrier via TCS-controlled fetchers: loader A holds the gate, waiter B is cancelled, last-leaver exit of A races a new registrant C → assert no second `loadAsync` starts while A is active and exactly one cache write lands.

### LINKED ISSUE #479: api: HandleAddSource logs source URLs with query-string tokens — violates the #149 log policy
## Found by
External audit 2026-09-04 (security pass over the management API).

## Summary
`HandleAddSource` logs the raw source URL three times through `SanitizeForLog` (`BGPLite.Api/ManagementApi.cs:1254`, `:1259`, `:1270`):

```csharp
_logger.LogWarning("Save-time URL validation timed out for '{Url}'", SanitizeForLog(data.Url));
_logger.LogWarning("Save-time URL validation rejected '{Url}': {Error}", SanitizeForLog(data.Url), validationError);
_logger.LogInformation("Added source '{Name}' ({Url}) to peer {PeerId}", ...);
```

`SanitizeForLog` strips control characters and truncates to 200 chars — **the query string survives**. A source `https://host/list?token=SECRET` writes the token into persistent Warning/Information application logs. The `validationError` string returned by `PrefixSourceUrlValidator` also embeds the URL and is logged verbatim at `:1259`.

This directly violates the #149 policy the codebase enforces everywhere else ("peer-supplied source URLs may carry query-string tokens — log the source name, never the URL"): `HttpPrefixProvider` and `UserSourceCache` log the source `Name` only, and `Md5Password` is never logged or echoed (verified by grep) — these three call sites are the outlier.

## Fix shape
Log only `data.Name` plus a short reason code at all three sites; keep the URL out of the message templates (returning it in the HTTP response body is fine — that is the operator's own input, not a log).

## Severity
MEDIUM — secrets written to persistent logs; reachability is one management-API call (loopback trust boundary by design, but the log files outlive it).

## Regression test
Following the `LogSanitizationTests` pattern: `HandleAddSource` with `?token=x` emits no event text containing `token=` (assert against the captured log records).

### LINKED ISSUE #480: api,providers: /api/asn-lists aborts the connection when the ExternalFetchBudget expires instead of returning the partial response
## Found by
External audit 2026-09-04 (API error-path pass). Related family: #424 (in-flight occupancy), #454 (budget on a different endpoint) — neither covers this abort-vs-partial behavior.

## Summary
`HandleAsnLists` awaits `_prefixSources.LoadAllAsync(ct)` unguarded (`BGPLite.Api/ManagementApi.cs:1443`). The per-source fetches inside `LoadAllAsync` degrade correctly on a foreign-token budget OCE (`PrefixSourceService.cs:129-132` — per-source warning + partial counts), but once the request-level ExternalFetchBudget (30 s, `ManagementApi.cs:70`) fires, `ct.IsCancellationRequested` is true, the `when` filter rethrows, and the OCE escapes the handler:

```text
LoadAllAsync throws OCE (caller token)
  → :1443 unguarded
  → HandleAsync rethrows OCE
  → HandleWithInflightReleaseAsync: the `when (_shutdownCts.IsCancellationRequested)` filter is FALSE (this was the budget, not shutdown)
  → generic catch → LogError + connection closed WITHOUT a body
```

The partial-JSON construction at `:1458-1461` is dead code on exactly the path it was written for. `HandleExportPrefixes:1330-1338` already implements the intended degrade-and-return pattern for the same class.

## Current behavior
A slow RIPE day turns `/api/asn-lists` into connection resets; the data collected before the budget expires is thrown away and the client gets no diagnostics.

## Fix shape
Wrap the call: `catch (OperationCanceledException) when (!_shutdownCts.IsCancellationRequested)` → warning log + return the partial result assembled so far (mirror `HandleExportPrefixes`).

## Severity
MEDIUM — API reliability under upstream slowness; no data corruption.

## Regression test
Fake source service: first source resolves, second blocks past a tiny budget → assert HTTP 200 with the first source present and the partial-degradation signal, not a dropped connection.

### LINKED ISSUE #481: config,server: MaxPrefixesPerPeer defaults to 0 (unlimited) — the RFC 4486 defense is off out of the box
## Found by
External audit 2026-09-04 (resource-exhaustion pass). Not a #304/#391 duplicate: those implemented the mechanism; this is about the shipped default leaving it off.

## Summary
The per-peer prefix ceiling implemented for #304/#391 is complete and correct (Cease/MaxPrefixesExceeded per RFC 4486 §2, per-peer overrides, 75 % warning) — but `BgpConfig.MaxPrefixesPerPeer` defaults to `0` = unlimited (`BGPLite.Configuration/BgpConfig.cs:93`), so a stock deployment enforces nothing:

```csharp
[YamlMember(Alias = "MaxPrefixesPerPeer")]
public int MaxPrefixesPerPeer { get; init; } = 0;
```

Combined with D11 (auto-registration: any Internet speaker that completes a valid OPEN becomes a persisted, served peer), one established "peer" can fill the shared table with unique IPv4 /24-/32 and IPv6 /48-/128 prefixes: each unique NLRI is a `RouteTable` entry + `Route` object + an `_installedPrefixes` entry — memory grows linearly with unique prefixes and nothing bounds it while the session is up. `MaxAcceptsPerIpPerMinute` bounds the connection rate, not the table; teardown cleans up, so this is an unbounded-while-connected vector, not a leak.

## Fix shape
Either:
- **(a)** a generous non-zero default (e.g. `1_000_000` — above any legitimate route-server peer, far below memory exhaustion), or
- **(b)** keep 0 for compatibility, but `config.Validate()` emits a startup Warning naming the exposure when the effective global cap is 0 and auto-registration is active.

(a) is the safer out-of-the-box posture; (b) is the minimal honest signal.

## Severity
MEDIUM — remote memory-pressure DoS by default; requires only an established session, which D11 hands to any BGP speaker on the Internet.

## Regression test
Config-level: assert the shipped default is non-zero (for (a)), or that `Validate` warns at 0 (for (b)); the enforcement behavior itself is already covered by the #304 tests.

### LINKED ISSUE #482: server: teardown dispose-race noise — CancelAsync ODE, unguarded initial-send lock, _cts.Token ODE, SendUpdateBatchAsync ignores the send result
## Found by
External audit 2026-09-04 (concurrency pass; CTS/semaphore semantics verified with runtime probes on .NET 10). One theme, four spots — a single hygiene PR closes them all. No protocol/routing invariant breaks; the cost is misleading ERROR logs on routine operations plus unobserved task faults.

## The four spots

1. **`RunEstablishedAsync` — `await _cts.CancelAsync()` throws `ObjectDisposedException` when an external `Dispose()` wins the race** (`BgpSession.cs:680`, and the hold-time-0 path at `:669`). Between `_cts.Cancel()` and `_cts.Dispose()` in `Dispose` (`:2001-2003`) sit two instructions; `TerminateSessionsAsync` (API peer deletion, `BgpServer.cs:516-523`) does `NotifyCeaseAsync` → `Dispose` while the session task is parked on `Task.WhenAny` (`:679`). The ODE lands in the generic catch (`:557`) → ERROR "Session error" with an ODE stack trace on a **routine** peer deletion, and `AwaitLoopTaskAsync` (`:682-683`) is skipped — the read/keepalive loop faults go unobserved. `MarkSilentClose:1943` and `FaultSession:226` already wrap the synchronous `Cancel()` in an ODE-catch; `CancelAsync` was missed. Fix: the same try/catch.

2. **Initial-send `_advertisedPrefixesLock` acquire/release unguarded** (`:481`, `:490`): Dispose during the route dump → OCE/ODE into the generic catch → ERROR noise. `RefreshCycleAsync:170-208` already shows the guarded form (acquire: OCE+ODE; release: ODE+SemaphoreFullException). Mirror it. (The cancel-before-dispose semaphore ordering inside `Dispose` is load-bearing and correct — probed: cancel-then-dispose wakes parked waiters with OCE; the reverse order would hang them forever.)

3. **`_cts.Token` read in `SendAllRoutesAsync` without the ODE guard** (`:1382`, `:1396`): a refresh outliving Dispose reads `.Token` on a disposed CTS → ODE surfaced as "Could not resolve the per-peer MaxPrefix override" (`:1386-1391`) or "Failed to refresh routes" (`RefreshCycleAsync:200`) — wrong diagnostics for a normal unwind. `RefreshRoutesAsync:126-128` already guards the identical read; normalize the token once and pass it down.

4. **`SendUpdateBatchAsync` discards `SendMessageAsync`'s `bool`** (`:1644`): in the Dispose window the send returns false (send-lock disposed / connection disposed) but the code still appends to `_advertisedPrefixes` (`:1649-1658`) and returns true → the mirror and `_advertisedCount` record prefixes that never reached the wire — a narrow hole in the #212/#457 wire-truth contract. No escalation (post-dispose nobody reads them; subsequent withdrawals go to a dead socket). Fix: `if (!await SendMessageAsync(update)) return false;` — safe because false there means dispose, not a truncated frame (#285 stays in `SocketBgpConnection`).

## Severity
LOW-MEDIUM — operational (misleading logs, unobserved faults); no protocol or routing impact.

## Regression test
Extend the #341 dispose-during-send pattern: Dispose while parked on `Task.WhenAny` → assert no ERROR-level event is logged and `RunAsync` completes normally.

### LINKED ISSUE #483: server,protocol: pre-OPEN phase replies 2/0 to a peer NOTIFICATION and classifies non-OPEN first messages as 2/0 instead of FSM 5/0
## Found by
External audit 2026-09-04 (FSM pass). Same class as #427 (Established) and #453 (OpenConfirm) — this is the third, pre-OPEN phase, and the last unhandled one.

## Summary
The first received message is handled at `BgpSession.cs:371-375`:

```csharp
if (openMessage is not BgpOpenMessage remoteOpen)
{
    await SendNotificationAsync(BgpConstants.Error.OpenMessageError, BgpConstants.SubError.Unspecific);
    return;
}
```

Two defects:

1. **A peer NOTIFICATION as the first message gets a reply.** RFC 4271 §6.3/§4.5 (and the project's own D8: "a peer NOTIFICATION is never replied to") require silent teardown on receiving a NOTIFICATION — BGPLite answers with 2/0 instead. The OpenConfirm branch right below (`:458-465`) gets this right (log + return, no reply); `ReadLoopAsync`'s Established arm (`:886-895`) also latches `RemoteNotification` correctly.
2. **KEEPALIVE/UPDATE/ROUTE_REFRESH as the first message is answered 2/0 (Open Message Error)** — but a non-OPEN message during the handshake phase is an FSM-level violation ("the handshake phase accepts only OPEN and NOTIFICATION" — the #453 rationale), answered 5/0 everywhere else. Inconsistent with #427 and #453.

Additionally, both pre-Established sends (`:373`, `:468`) bypass the `_teardownReason` CAS latch. The single-NOTIFICATION invariant still holds (all senders enumerated: no external sender can overlap a pre-Established session — Cease loops filter `IsEstablished`, the hold-timer loop is not running yet), but a session being replaced (`MarkSilentClose`, `BgpServer.cs:336`) can still emit a NOTIFICATION from these two sites, defeating the silent-close intent.

## Fix shape
- First message is NOTIFICATION → log + latch `RemoteNotification` + return (no reply);
- other non-OPEN first messages → CAS-latch `LocalCease` + NOTIFICATION 5/0 (Finite State Machine Error);
- add the same CAS to the OpenConfirm default branch (`:466-469`).

## Severity
LOW — interop edge (a peer must send a message before our OPEN — e.g. immediate-reject speakers); the wire consequence is one wrong-shaped NOTIFICATION on an already-dying connection.

## Regression test
ScriptedConnection: first frame = NOTIFICATION(6,0) → assert zero outbound NOTIFICATION bytes; first frame = KEEPALIVE → assert exactly one NOTIFICATION 5/0 and teardown.

### LINKED ISSUE #484: protocol,server: treat-as-withdraw skips the MP_UNREACH_NLRI carried by the same UPDATE
## Found by
External audit 2026-09-04 (UPDATE-semantics pass). Not covered by #467 (that settled the §3(g)/§3(j) MP error policy) — this is an ordering asymmetry inside the normal treat-as-withdraw path.

## Summary
The classic withdrawn routes of an UPDATE are applied BEFORE attribute parsing (`BgpSession.cs:1038-1039`), so they survive an attribute-pipeline failure. The MP_UNREACH half is processed AFTER the announcement block (`:1264-1276`) and is therefore lost when `ParseRouteAttributes` throws: the treat-as-withdraw catch (`:1065-1092`) withdraws the classic NLRI and MP_REACH prefixes of the message and rethrows — the MP_UNREACH prefixes named in the SAME UPDATE are never withdrawn.

```text
UPDATE [classic NLRI + malformed ORIGIN + MP_UNREACH(v6: 2001:db8::/32 x)]
  → classic NLRI correctly treated-as-withdrawn (:1082-1083)
  → MP_REACH prefixes withdrawn (:1084-1089)
  → MP_UNREACH half skipped (its block at :1264-1276 is never reached after the rethrow)
  → the stale IPv6 route lives until session teardown or a repeat withdrawal
```

RFC 7606 §2 treats the message as a whole ("all contained routes … as though withdrawn"); the codebase's own ordering for the classic WITHDRAWN ROUTES field (applied before the parse, precisely so it survives) shows the intent — the asymmetry is an oversight, not a decision (nothing in D17/D2 records it).

## Fix shape
In the treat-as-withdraw catch, also `WithdrawIfOwned` every prefix of `update.MpUnreachV6` / `update.MpUnreachV4` before the rethrow (same ownership rule as everywhere else).

## Severity
LOW — requires a mixed UPDATE (announcements + MP-withdrawal + a malformed attribute); the stale route is bounded by session teardown.

## Regression test
Install an IPv6 route via a valid MP_REACH → send UPDATE [some classic NLRI + malformed ORIGIN + MP_UNREACH naming the v6 prefix] → assert the v6 route is gone from the table.

### LINKED ISSUE #485: providers: unfiltered OCE rethrows — RipeStatPrefixCache bypasses stale/negative handling, PrefixService RU path rethrows budget OCE, WarmUp swallows caller cancellation
## Found by
External audit 2026-09-04 (cancellation-semantics pass). One pattern, three sites — all diverge from the repo's own #114/#225/#320/#324 contract: caller-OCE propagates; foreign-OCE (budget/body-deadline with a live caller token) is a load FAILURE → stale-serve / negative-cache.

## Sites

1. **`RipeStatPrefixCache.cs:137`** — `catch (OperationCanceledException) { throw; }` without `when (ct.IsCancellationRequested)`: a body-deadline OCE (foreign token, live ct, `RipeStatProvider.cs:47-67`) escapes the stale-serve/negative-cache handling below → every subsequent call re-pays the full fetch budget; no backoff is ever recorded. The reference implementations of the same contract are `UserSourceCache.cs:127-135` and `PrefixSourceService.cs:129-132`.
2. **`PrefixService.cs:253`** (`GetRuPrefixesAsync`) — the same unfiltered rethrow: a #324 budget OCE from the default-source load propagates as "cancellation" while the caller token is alive; consumers currently mask it in generic catches, but it is the latent #225 pattern one refactor away from surfacing.
3. **`PrefixService.cs:327`** (`WarmUpAsync` per-ASN loop) — the INVERSE defect: `catch (Exception ex)` swallows the host-shutdown OCE per ASN → a "WarmUp failed" WARN for every remaining ASN on every shutdown, instead of unwinding.

## Fix shape
- (1) + (2): `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` — the foreign OCE then falls through to the stale/negative branch, byte-identical to the two reference sites.
- (3): add `catch (OperationCanceledException) { throw; }` before the generic catch inside the loop.

## Severity
LOW — contract hygiene + log noise + a lost backoff window; no data corruption (stale copies still serve where they exist).

## Regression test
Following the `PrefixCacheFailureTests` pattern: (1) body-deadline OCE → assert the negative entry is recorded; (2) budget OCE on the default source → assert the stale copy is served; (3) WarmUp with a cancelled token → assert no per-ASN WARN storm (first OCE unwinds the loop).

### LINKED ISSUE #486: protocol: a zero-length AS_PATH is accepted as valid inbound — settle the policy

### LINKED ISSUE #487: chore: minor-batch from the 2026-09-04 audit — statusTimer dispose, hardcoded 179, watcher errors, GET→UPDATE TOCTOU, file cap, AddSource limits, O(n²) eviction, stale comments
## Found by
External audit 2026-09-04. Verified minor items that do not warrant individual issues; one sweep PR can take them all (same pattern as #238/#321 batch issues).

- **`BgpServer.Dispose` never disposes `_statusTimer`** (`BgpServer.cs:570-578`; `StopAsync` does at `:163`) — the `PeriodicTimer` lives until GC on the host-abort path.
- **Hardcoded `179`** at `ManagementApi.cs:724` instead of `BgpConstants.BgpPort` (AGENTS.md: "Do not hardcode either port — reference the constant/config").
- **`ConfigReloader` does not subscribe `FileSystemWatcher.Error`** (`ConfigReloader.cs:63-74`): an internal-buffer overflow (busy save storms) silently drops change events until the next save — hot reload quietly stops working.
- **GET→UPDATE TOCTOU against a concurrent DELETE** (`ManagementApi.cs:1057→1116`): `ExecuteUpdate` affects 0 rows, the handler still logs "Updated peer", and the client receives a 404 from the follow-up GET. Check `rowsAffected` and answer 404 directly.
- **`FilePrefixProvider.cs:44` reads the file with no size cap** (`File.ReadAllTextAsync`; every HTTP path caps at 10 MB). Operator-controlled input, so the amplification is config-only — but the asymmetry invites a future mistake.
- **`HandleAddSource`/`PeerStore.AddCustomSourceAsync` have no field limits** — no max length for `Name`/`Url`, no cap on the number of sources per peer. The 1 MiB body cap bounds one request only; DB growth is unbounded (loopback trust boundary, but peer rows are Internet-reachable via D11 auto-registration + operator flows).
- **`UserSourceCache` eviction is O(n²)** — `toEvict.Contains` on a `List<string>` at `:213`/`:276` (up to 1024 entries); use a `HashSet`.
- **Stale comment** — `UpdateCodec.cs:313-315` still says the MP_REACH/MP_UNREACH duplicate handling "is deliberately not implemented … belongs with MP-BGP support (#14)"; #467 implemented it (in `BgpMessageReader.ParseUpdate`). The comment now misdirects.
- **`_remoteAsn` / `_peerMpIpv6Unicast` are not volatile** (`BgpSession.cs:35,75`) while read cross-thread (BgpServer session filters `BgpServer.cs:437/:471`, the send path `BgpSession.cs:1443`); `_peerMpV6Disabled` is volatile — inconsistent. Practical visibility is immediate via thread-pool barriers; fix opportunistically with `Volatile.Read/Write`.

## Severity
LOW each — hygiene, observability, hardening.

### LINKED ISSUE #488: routing: settle seed-takeover loss on session teardown and the total-source-failure RU fallback (product decisions)
## Found by
External audit 2026-09-04 (routing pass). Both are intentional-adjacent behaviors that deserve an explicit decision + D-entry rather than silence (same shape as #457: "settle the policy").

## 1. Seed-takeover loss on session teardown
`RouteSeedingService` installs startup prefixes owner-less (`RouteSeedingService.cs:85`); a peer announcing the same prefix takes over the entry (D12, "last writer wins" — documented for REPLACEMENT); when that session ends, `RemoveAllOwnedBy(this)` (`BgpSession.cs:598`) deletes the entry and nothing restores the seed value.

- Production composition: only `GET /api/routes` / LPM lose the prefix (the outbound path reads sources, not the table).
- Degraded composition (`SharedTableRouteAssembler`, D13): the prefix stops being advertised to everyone until a process restart.

D12 covers "the loser cannot withdraw what it no longer owns" — it does not record that the seed's VALUE (not just ownership) disappears with the taker's teardown. Decide: restore the seed value on ownership loss (requires keeping the seed data separately), or record the loss as accepted.

## 2. Total-source-failure RU fallback
`RouteAssembler.cs:241-255`: a configured peer whose sources ALL resolve to 0 prefixes — **including the case where every fetch FAILED** (each failure is individually caught and logged, then counted as empty) — falls back to the full RU default list. During a RIPEstat/network outage every subscribed peer receives the entire RU table it never asked for (initial send after reconnect included).

Decide: distinguish "source returned 0" from "source fetch failed" and do not substitute on total failure; or record the fail-open fallback as intended behavior. Related nuance in the same area: `RouteAssembler.cs:155-159` silently ignores a subscription name that matches neither `RipeStat.AsnLists` nor `PrefixSources` — a config typo is invisible (no log).

## Severity
LOW — product/policy; both behaviors are logged today, neither corrupts state.

## Regression test
After the decision: (1) seed takeover + teardown → the seed is visible again (or pinned as accepted loss); (2) all-sources-failed → the peer keeps its previous/empty set, not the RU dump (or pinned as intended); (3) an unknown subscription name produces at least one startup warning.

ruhex:
Integration of the 2026-09-04 audit batch (tracking #489). One PR per issue, all CI-green and CodeQL-clean on their dev merge-refs.

Closes #476
Closes #477
Closes #478
Closes #479
Closes #480
Closes #481
Closes #482
Closes #483
Closes #484
Closes #485
Closes #486
Closes #487
Closes #488

## Per-issue summary
- **#476** (PR #490, HIGH): `MergeDuplicatePrefixes` now carries `IsIpv4` — a duplicated IPv6 prefix no longer crashes the initial send (flap loop) or leaks a bogus `0.0.0.0/N` classic NLRI. One-line fix + red/green test.
- **#477** (PR #491): the documented-valid `PrefixSources:` YAML-null no longer NREs at startup — `?? []` at all seven consumers.
- **#478** (PR #492): UserSourceCache gate bookkeeping (register / last-leaver removal / sweeps) is one atomic section under `_gateSync` — the #468 duplicate-fetch TOCTOU window closed by construction.
- **#479** (PR #493): source URLs with query-string tokens stay out of the logs (#149 policy); the validator's reason moved to the 400 response.
- **#480** (PR #494): `/api/asn-lists` degrades to the partial JSON on ExternalFetchBudget expiry instead of aborting the connection.
- **#481** (PR #495): `MaxPrefixesPerPeer` ships bounded at 1,000,000 (RFC 4486 defense on out of the box; 0 = explicit opt-out). **Behavior change.**
- **#482** (PR #496): four teardown dispose-race guards (CancelAsync ODE, initial-send lock, `_cts.Token` reads, send-mirror wire-truth).
- **#483** (PR #497): pre-OPEN phase per the FSM — a peer NOTIFICATION is never replied to, other non-OPEN first messages get exactly one 5/0; both sends CAS-latched. **Wire-visible behavior change** for peers that talk before our OPEN.
- **#484** (PR #498): treat-as-withdraw applies the MP_UNREACH_NLRI of the same UPDATE (RFC 7606 §2 whole-message semantics).
- **#485** (PR #499): caller-vs-foreign OCE contract at RipeStatPrefixCache / PrefixService RU path; WarmUp unwinds on shutdown instead of a WARN per ASN.
- **#486** (PR #500): zero-length AS_PATH rejected at the eBGP policy layer (3/11, codec stays encoding-neutral); recorded as D25. **Behavior change.**
- **#487** (PR #501): minor batch — statusTimer dispose, BgpConstants.BgpPort, FileSystemWatcher.Error, PATCH-vs-DELETE 404, file-size cap parity, AddSource field/count limits, O(n²) eviction, stale comment, volatile fields.
- **#488** (PR #502): total-source-failure fails CLOSED (no RU dump substitution), unknown subscription names warned, both recorded as D26. **Behavior change.**

## Verification
- Per-PR: `dotnet format` / `dotnet build BGPLite.sln` 0w/0e / full `dotnet test` green; dev CI (build-test, format-check, analyze) green; CodeQL 0 alerts on every merge-ref.
- Cumulative suite: 1074 → **1092** tests (15 red/green-proven regressions).
- Design decisions recorded: D25, D26; DESIGN_DECISIONS updated.

## Summary by CodeRabbit

- **New Features**
  - Added configurable per-peer prefix protection, enabled by default with a 1 million-prefix limit.
  - Added limits for custom source names, URLs, and the number of sources per peer.
  - Added warnings when configuration changes cannot be monitored or subscriptions reference unknown sources.

- **Bug Fixes**
  - Improved handling of source outages, cancellations, oversized files, empty configurations, and concurrent peer changes.
  - Prevented sensitive source URL tokens from appearing in logs.
  - Improved BGP protocol error handling, route withdrawals, IPv6 announcements, and session shutdown reliability.

- **Documentation**
  - Documented the prefix limit and updated design decisions for routing and protocol behavior.

### DIFF
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1445,20 +1445,31 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         // Append configured PrefixSources (file/http) alongside the legacy RipeStat ASN-lists,
         // reusing the same response shape. "Kind" is intentionally not exposed.
+        // #480: the budget token firing mid-LoadAllAsync surfaces as a foreign-token OCE (live
+        // shutdown token) — without this catch it escaped the handler and closed the connection
+        // without a body, discarding the partial result assembled above. Degrade like the
+        // per-source loops above; the tail warning below reports the budget expiry (both paths).
         var seen = lists.Select(l => l.Name).ToHashSet();
-        foreach (var (source, prefixes) in await _prefixSources.LoadAllAsync(ct))
+        try
         {
-            if (!seen.Add(source.Name)) continue; // skip names already present (e.g. shared "ru")
-            result.Add(new
+            foreach (var (source, prefixes) in await _prefixSources.LoadAllAsync(ct))
             {
-                id = source.Name,
-                Name = source.Name,
-                Description = source.Description,
-                Country = (string?)null,
-                Community = source.Community,
-                prefixCount = prefixes.Count,
-                type = source.Kind == "asn" ? "asn" : "list"
-            });
+                if (!seen.Add(source.Name)) continue; // skip names already present (e.g. shared "ru")
+                result.Add(new
+                {
+                    id = source.Name,
+                    Name = source.Name,
+                    Description = source.Description,
+                    Country = (string?)null,
+                    Community = source.Community,
+                    prefixCount = prefixes.Count,
+                    type = source.Kind == "asn" ? "asn" : "list"
+                });
+            }
+        }
+        catch (OperationCanceledException) when (!_shutdownCts.IsCancellationRequested)
+        {
+            // Budget expiry, not shutdown — serve the partial result (see above).
         }
 
         if (ct.IsCancellationRequested && !_shutdownCts.IsCancellationRequested)
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -68,6 +68,12 @@ public sealed class ManagementApi : IHostedService, IDisposable
     /// their PARTIAL result with a warning; shutdown still propagates. internal settable for tests.
     /// </summary>
     internal TimeSpan ExternalFetchBudget { get; set; } = TimeSpan.FromSeconds(30);
+
+    // #487: bounds for peer-supplied custom sources — generous ceilings that only reject abuse
+    // (repeated POSTs growing the DB without limit; megabyte-scale names/URLs in logs and rows).
+    internal const int MaxSourceNameLength = 200;
+    internal const int MaxSourceUrlLength = 2048;
+    internal const int MaxSourcesPerPeer = 64;
     // #258: in-flight tracking is a COUNT plus an idle Task (completed whenever the count is 0),
     // not a List<Task> — the #248 design appended every handler and never removed it, so each
     // completed request leaked its Task (and its exception, if faulted) for the process lifetime
@@ -721,7 +727,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         {
             asn = bgp.Asn,
             routerId = bgp.RouterId,
-            bgpPort = 179,
+            bgpPort = BgpConstants.BgpPort,
             apiPort = _port,
             holdTime = bgp.HoldTime,
             keepalive = bgp.KeepAlive,
@@ -1115,13 +1121,18 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         await _store.UpdatePeerConfigurationAsync(peerId, data.Description, data.Lists, parsedPrefixes, data.CustomAsns, data.MaxPrefix,
             md5Password: data.Md5Password);
+        // #487: a concurrent DELETE between the top-of-handler GET and this UPDATE left the
+        // ExecuteUpdates touching 0 rows while the handler still logged "Updated peer", answered
+        // 200 and fired a refresh — the follow-up GET then 404ed. Re-check the row and answer 404
+        // directly (the write itself is harmless: transaction-scoped, 0 rows affected).
+        var surviving = await _store.GetDbPeerByIdAsync(peerId);
+        if (surviving is null)
+            return ApiResponse.Error("Peer not found", 404);
         if (data.Md5Password is not null)
         {
             // #418: re-arm from ALL surviving rows for this IP — clearing this peer's key must
             // fall back to a sibling's, not silently disarm the address for every peer on it.
-            var md5Peer = await _store.GetDbPeerByIdAsync(peerId);
-            if (md5Peer is not null)
-                await RearmPeerIpMd5KeyAsync(md5Peer.Ip);
+            await RearmPeerIpMd5KeyAsync(surviving.Ip);
         }
 
         _logger.LogInformation("Updated peer {Id}", SanitizeForLog(peerId));
@@ -1216,6 +1227,16 @@ public sealed class ManagementApi : IHostedService, IDisposable
 
         if (data is null || string.IsNullOrWhiteSpace(data.Name) || string.IsNullOrWhiteSpace(data.Url))
             return ApiResponse.Error("Name and Url are required", 400);
+
+        // #487: bound the peer-supplied fields and the per-peer source count. The 1 MiB body cap
+        // bounds ONE request; nothing bounded repeated posts, and the DB rows outlive the request.
+        // Generous limits: a longer name/URL is never legitimate, and a peer needs far fewer
+        // sources than the user-source cache's 1024-URL entry cap.
+        if (data.Name.Length > MaxSourceNameLength || data.Url.Length > MaxSourceUrlLength)
+            return ApiResponse.Error(
+                $"Name must be at most {MaxSourceNameLength} characters and Url at most {MaxSourceUrlLength}.", 400);
+        if ((await _store.GetCustomSourcesAsync(peerId)).Count >= MaxSourcesPerPeer)
+            return ApiResponse.Error($"Peer already has the maximum of {MaxSourcesPerPeer} custom sources.", 400);
 
         // #266 item 3: the community is peer-supplied and reaches the wire through
         // CommunityCodec at send time — validate the SAME contract at the API boundary (#255
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1250,14 +1250,19 @@ public sealed class ManagementApi : IHostedService, IDisposable
         }
         catch (OperationCanceledException) when (validationCts.IsCancellationRequested)
         {
-            // DNS-resolution timeout (5s).
-            _logger.LogWarning("Save-time URL validation timed out for '{Url}'", SanitizeForLog(data.Url));
+            // DNS-resolution timeout (5s). #479 (#149): source URLs may carry query-string tokens —
+            // the log names the source, never the URL.
+            _logger.LogWarning("Save-time URL validation timed out for source '{Name}'", SanitizeForLog(data.Name));
             return ApiResponse.Error("URL validation timed out (DNS resolution took too long)", 400);
         }
         if (!isValid)
         {
-            _logger.LogWarning("Save-time URL validation rejected '{Url}': {Error}", SanitizeForLog(data.Url), validationError);
-            return ApiResponse.Error($"Invalid URL: the host could not be reached or is not allowed", 400);
+            // #479 (#149): neither the URL nor the validator's message (which embeds it) belongs in
+            // the log. The reason goes to the RESPONSE instead — the operator just submitted this
+            // URL, so echoing it there leaks nothing.
+            _logger.LogWarning("Save-time URL validation rejected source '{Name}' for peer {PeerId}",
+                SanitizeForLog(data.Name), SanitizeForLog(peerId));
+            return ApiResponse.Error($"Invalid URL: {validationError ?? "the host could not be reached or is not allowed"}", 400);
         }
 
         var source = await _store.AddCustomSourceAsync(peerId, data.Name, data.Url, data.Community);
@@ -1266,8 +1271,9 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // same pattern as CreatePeer/UpdatePeer. Pass ASN so shared-IP peers aren't refreshed (#200).
         RequestPeerRefresh(peerId, peer);
 
-        _logger.LogInformation("Added source '{Name}' ({Url}) to peer {PeerId}",
-            SanitizeForLog(data.Name), SanitizeForLog(data.Url), SanitizeForLog(peerId));
+        // #479 (#149): log the source Name, not the URL (query-string tokens are secrets).
+        _logger.LogInformation("Added source '{Name}' to peer {PeerId}",
+            SanitizeForLog(data.Name), SanitizeForLog(peerId));
         return ApiResponse.Ok(new { id = source.Id, name = source.Name, url = source.Url, community = source.Community, active = source.Active });
     }
 
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1712,7 +1712,7 @@ public sealed class ManagementApi : IHostedService, IDisposable
         if (config.RipeStat?.AsnLists is { } lists)
             foreach (var l in lists)
                 known.Add(l.Name);
-        foreach (var source in config.PrefixSources)
+        foreach (var source in config.PrefixSources ?? []) // #477: YAML null = no sources
             known.Add(source.Name);
         return names.Where(n => !known.Contains(n)).Distinct(StringComparer.Ordinal).ToList();
     }
--- a/BGPLite.Api/ManagementApi.cs
+++ b/BGPLite.Api/ManagementApi.cs
@@ -1264,10 +1264,11 @@ public sealed class ManagementApi : IHostedService, IDisposable
         // (unlike ASP.NET Core's RequestAborted), so the timeout is purely time-bounded.
         using var validationCts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
         bool isValid;
-        string? validationError;
         try
         {
-            (isValid, validationError) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url, ct: validationCts.Token);
+            // #503 review: the validator's reason string embeds the URL and DNS internals —
+            // discarded; the log names the source and the response carries a fixed classification.
+            (isValid, _) = await PrefixSourceUrlValidator.ValidateUrlAsync(data.Url, ct: validationCts.Token);
         }
         catch (OperationCanceledException) when (validationCts.IsCancellationRequested)
         {
@@ -1279,11 +1280,13 @@ public sealed class ManagementApi : IHostedService, IDisposable
         if (!isValid)
         {
             // #479 (#149): neither the URL nor the validator's message (which embeds it) belongs in
-            // the log. The reason goes to the RESPONSE instead — the operator just submitted this
-            // URL, so echoing it there leaks nothing.
+            // the log. #503 review (CWE-209): the validator's message ALSO embeds resolved
+            // addresses and raw DNS exception text — classify for the response instead of echoing
+            // server-side resolution internals back to the caller.
             _logger.LogWarning("Save-time URL validation rejected source '{Name}' for peer {PeerId}",
                 SanitizeForLog(data.Name), SanitizeForLog(peerId));
-            return ApiResponse.Error($"Invalid URL: {validationError ?? "the host could not be reached or is not allowed"}", 400);
+            return ApiResponse.Error(
+                "Invalid URL: the scheme must be http/https on port 80/443 and the host must resolve to a reachable public address", 400);
         }
 
         var source = await _store.AddCustomSourceAsync(peerId, data.Name, data.Url, data.Community);
--- a/BGPLite.Configuration/BgpConfig.cs
+++ b/BGPLite.Configuration/BgpConfig.cs
@@ -88,9 +88,12 @@ public sealed class BgpConfig
     /// <summary>#304: per-peer ceiling on prefixes installed from one session (distinct
     /// NLRI this session currently owns); exceeding it tears the session down with
     /// NOTIFICATION(Cease, MaxPrefixesExceeded) per RFC 4271 §6.7 / RFC 4486 §2.
-    /// 0 = unlimited (the convention of OpenTimeoutSeconds / MaxAcceptsPerIpPerMinute).</summary>
+    /// #481: the shipped default is bounded (1,000,000 — above any legitimate provisioning
+    /// peer, far below memory exhaustion) so the RFC 4486 defense is on out of the box;
+    /// 0 = unlimited remains available as an explicit opt-out (the convention of
+    /// OpenTimeoutSeconds / MaxAcceptsPerIpPerMinute).</summary>
     [YamlMember(Alias = "MaxPrefixesPerPeer")]
-    public int MaxPrefixesPerPeer { get; init; } = 0;
+    public int MaxPrefixesPerPeer { get; init; } = 1_000_000;
 
     public IPAddress GetRouterIdAddress() => IPAddress.Parse(RouterId);
 
--- a/BGPLite.Protocol/UpdateCodec.cs
+++ b/BGPLite.Protocol/UpdateCodec.cs
@@ -310,9 +310,11 @@ public static class UpdateCodec
             // one, so anything reading the first (a collector, a looking glass, an operator's packet
             // capture) disagreed with what actually landed in the route table (#287).
             //
-            // The MP_REACH_NLRI/MP_UNREACH_NLRI half of that paragraph — duplicate → NOTIFICATION,
-            // session reset — is deliberately not implemented: BGPLite is IPv4-unicast only and never
-            // parses those attributes. It belongs with MP-BGP support (#14).
+            // The MP_REACH_NLRI/MP_UNREACH_NLRI half of that paragraph is handled UPSTREAM, in
+            // BgpMessageReader.ParseUpdate: those attributes are extracted into typed fields
+            // before this generic list is built, and a duplicate takes the RFC 7606 §3(g)
+            // NOTIFICATION + session reset there (#467). The first-wins guard below therefore
+            // never sees them.
             //
             // Clear() is not redundant despite `localsinit` zeroing stackalloc by default: adding
             // [SkipLocalsInit] (or <SkipLocalsInit> in the csproj) is an ordinary perf tweak for a
--- a/BGPLite.Protocol/UpdateCodec.cs
+++ b/BGPLite.Protocol/UpdateCodec.cs
@@ -358,6 +358,16 @@ public static class UpdateCodec
                         originSeen = true;
                         break;
                     case BgpConstants.Attribute.AsPath:
+                        // #486 (D25): a zero-length AS_PATH is a legal ENCODING (RFC 4271 §5.1.2
+                        // lets an originator send an empty path toward INTERNAL peers) but never a
+                        // valid eBGP path — BGPLite serves eBGP only, where the sender's own ASN
+                        // must be present. Rejected at the policy layer as Malformed AS_PATH
+                        // (treat-as-withdraw); the codec stays encoding-neutral so the writer's
+                        // empty-path roundtrip (#248 review) stands.
+                        if (attr.Data.Length == 0)
+                            throw new BgpNotificationException(
+                                BgpConstants.Error.UpdateMessageError, BgpConstants.SubError.MalformedAsPath,
+                                "Invalid AS_PATH: an empty path is only valid toward internal peers (RFC 4271 §5.1.2)");
                         asPath = AttributeHelper.ReadAsPath(attr, fourByteAsnSession);
                         asPathSeen = true;
                         break;
--- a/BGPLite.Providers/FilePrefixProvider.cs
+++ b/BGPLite.Providers/FilePrefixProvider.cs
@@ -41,14 +41,29 @@ public sealed class FilePrefixProvider(ILogger<FilePrefixProvider> logger) : IPr
         // #321 item 5: async read with the caller's token — the sync ReadAllText held a threadpool
         // thread under the per-source gate for the whole file. The metadata probes above stay
         // sync: they are single stat calls, not reads.
-        // #487: cap parity with the HTTP paths (HttpPrefixProvider.MaxResponseBytes) — the file is
-        // operator-controlled, but a multi-GB "prefix list" is never legitimate and the uncapped
-        // read was an asymmetric memory-amplification footgun.
-        var length = new FileInfo(fullPath).Length;
-        if (length > HttpPrefixProvider.MaxResponseBytes)
-            throw new InvalidOperationException(
-                $"Prefix file for source '{source.Name}' is {length} bytes — over the {HttpPrefixProvider.MaxResponseBytes}-byte cap.");
-        var text = await File.ReadAllTextAsync(fullPath, ct);
+        // #487 + #503 review: cap parity with the HTTP paths (HttpPrefixProvider.MaxResponseBytes),
+        // enforced WHILE reading — a length probe alone races a concurrent writer extending the
+        // file between the stat and the read.
+        string text;
+        await using (var fs = File.OpenRead(fullPath))
+        {
+            if (fs.Length > HttpPrefixProvider.MaxResponseBytes)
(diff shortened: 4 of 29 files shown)


ruhex:
CI note: the first build-test run failed on `PrefixAutoRefreshServiceTests.PollsConditionalSourceMoreOftenThanNoEtagSource` (timing assertion, 441 ms) — second distinct CI timing flake this batch (after `RefreshDebounceTests` on #500, noted there). Re-run green with zero changes; both are poll/debounce tests sensitive to runner load. Recording as a pattern; no code change in this PR (one issue : one PR, and neither failure implicates the batch's diff).

ruhex:
## Review verdicts (per finding, verified against the code)

**Fixed via #504** (follow-up PR into dev, now part of this integration):
- ✅ WarmUp foreign-OCE filter (PrefixService) — valid, fixed.
- ✅ #488 gate semantics — valid divergence from the recorded D26 decision (code said ANY-failure, D26 says TOTAL); counters implemented, mixed-case test added, D26 wording aligned.
- ✅ FilePrefixProvider bounded read — valid TOCTOU on the length probe; cap now enforced during the read.
- ✅ CWE-209 validation error — valid (the validator's message embeds resolved addresses and raw DNS exception text, PrefixSourceUrlValidator.cs:93); fixed classification in the response.

**Dismissed with rationale:**
- 🚫 *Per-peer source cap atomicity* (ManagementApi): the management API is deliberately loopback-unauthenticated (D-decision, AGENTS.md security section) — the trust boundary is the operator's own machine, and the cap (#487) bounds operator mistakes and runaway scripts, not adversarial concurrency. A check-then-insert race between two concurrent loopback POSTs yielding 65-of-64 sources is not a reachable integrity violation in that boundary; true atomicity needs app-level serialization or a schema constraint. Recorded as future hardening *if* the API is ever exposed beyond loopback (that is itself a recorded product decision).
- 🚫 *RipeStat stale-serve records no failure state* (Minor): this is the uniform pre-existing stale-on-failure semantic — EVERY fetch failure that has a stale copy serves it without touching cache state (by design: serving stale IS the remedy; negative backoff applies only when no stale exists). The PR routes foreign-OCE into that same path rather than inventing a third behavior. Freshening CachedAt on stale-serve would mask real staleness; the next-caller refetch is the documented tradeoff of the whole stale-on-failure family (#163).
- 🚫 *Decouple route operations from _advertisedPrefixesLock* (heavy lift): the serialization is deliberate — the lock guards the send/withdraw MIRROR's consistency (build-then-commit, #430/#450), and the in-code comment documents why _sendLock cannot span the withdraw+re-announce pair (deadlock with the teardown NOTIFICATION paths). The wait is bounded by the per-send budget (#160/#285) and the refresh token. A redesign (event-sourced mirror / channel-based sender) is legitimate future work but is not a regression of this batch and does not belong in an integration PR.

ruhex:
@coderabbitai review

ruhex:
@coderabbitai review

ruhex:
CodeRabbit re-review of the follow-up commit (9820531) was attempted twice and is rate-limited (hourly quota). Review state: the full review ran on 5b2d857 (covered head), all 7 findings are dispositioned — 4 fixed via #504 (verified: CI + CodeQL + full suite green), 3 dismissed with rationale above. The follow-up commit contains exactly those four fixes and nothing else.

PR #511, opened 2026-09-04
chore: integrate dev — Cease subcode fixes #505–#506, timing-test determinization #507

### LINKED ISSUE #505: protocol,server: the per-peer MaxPrefixes reset loses its subcode — Cease 6/0 instead of 6/1 (BgpNotificationException swallowed by AwaitLoopTaskAsync on HoldTime>0 sessions)
## Found by
Local integration stand on `main @ ee1f73c` (post-#503): scripted BGP peer over real TCP through a userland sniffing proxy, wire frames decoded independently of BGPLite's own stack. Both findings reproduced deterministically; root causes located in source.

## Finding 1 (HIGH): the per-peer MaxPrefixes reset loses its subcode — Cease 6/0 instead of 6/1

**Wire evidence:** peer PATCHes `maxPrefix=3`, announces 4 prefixes → exactly one NOTIFICATION arrives, **code=6 subcode=0**. Expected **6/1** ("Maximum Number of Prefixes Reached", RFC 4486 §2/§3 — the RFC's prescribed handling for the #304/#391 cap is a MUST).

**Root cause chain:**
```
HandleUpdateAsync cap check throws BgpNotificationException(Cease, CeaseMaxPrefixes=1)
  → ReadLoopAsync does not catch it (correct — not an UPDATE body error)
  → RunEstablishedAsync: readTask faults; Task.WhenAny completes
  → AwaitLoopTaskAsync(readTask, "read") — its generic catch (Exception) LOGS
    "read loop faulted ... exceeded the per-peer prefix limit" and SWALLOWS the exception
    (it rethrows only BgpParseException, per #223)
  → RunEstablishedAsync returns normally → RunAsync finally: teardownReason==None →
    best-effort generic Cease **6/0**
```
The cap-throw's own comment (`BgpSession.cs`, the `#304` block) says it "unwinds to RunAsync's BgpNotificationException handler" — that intent broke when reads moved behind `Task.WhenAny`/`AwaitLoopTaskAsync`. The HoldTime==0 path still works correctly (`await ReadLoopAsync(ct)` propagates directly → 6/1), so the defect is specific to HoldTime>0 sessions.

**Impact:** the peer receives an unspecified Cease instead of the MaxPrefixes diagnostic — RFC 4486 §4 expects that subcode to tell the operator exactly what to fix; without it, automation/peers cannot distinguish "back off the prefix count" from a generic reset, inviting a reconnect loop against the cap. Functional reset/flush themselves work.

**Fix shape:** `AwaitLoopTaskAsync` must rethrow `BgpNotificationException` (alongside its `BgpParseException` rethrow) so `RunAsync`'s handler sends the carried code/subcode; the finally-Cease latch already suppresses the duplicate. Regression: hold-time>0 session at a MaxPrefix cap → sniffed NOTIFICATION is exactly one 6/1.

## Finding 2 (MEDIUM, one constant): CeaseAdministrativeReset is 6 — RFC 4486 assigns 4

**Wire evidence:** API peer deletion (`NotifyCeaseAsync`) → NOTIFICATION **6/6** arrives. RFC 4486 §3 table (verified against the RFC text): 4 = "Administrative Reset", 6 = "Other Configuration Change". `BgpConstants.SubError.CeaseAdministrativeReset = 6` (`BgpConstants.cs:74`) — every graceful reset/shutdown/delete is mislabeled as a configuration change. The neighboring constants (1/2/3) are correct per the same table.

**Fix shape:** constant 6 → 4 (+ update the delete/shutdown call-site docs if they quote the number). Regression: the same delete-peer stand scenario asserts 6/4.

## Severity
F1 HIGH (RFC 4486 MUST violation, reproducible, hold>0 sessions only). F2 MEDIUM (diagnostic mislabel on a correct teardown).

## Notes
- Both are pre-existing (not introduced by the #476–#488 batch); found by the integration stand's independent frame decoder, cross-checked against `host.log` ("read loop faulted … exceeded the per-peer prefix limit" + "NotificationSent … 6/0").
- Historical relative: #70 (closed) fixed blanket subcode-0 Ceases; F1 is the same wire symptom reintroduced by the loop-refactor swallow.
- The local stand (sniffing proxy + scripted peer + analyzer) is reusable: `/tmp/bgpit/` (host composition on ephemeral ports, since :179/tcpdump need root on macOS).

### LINKED ISSUE #506: protocol: CeaseAdministrativeReset constant is 6 — RFC 4486 §3 assigns 4 (Administrative Reset); 6 is Other Configuration Change
## Found by
Local integration stand on `main @ ee1f73c` (scripted BGP peer through a sniffing proxy, frames decoded independently); RFC fact-checked against the RFC 4486 §3 text. Extracted from #505 for 1 issue : 1 PR tracking (finding 2 there).

## Summary
`BgpConstants.SubError.CeaseAdministrativeReset = 6` (`BGPLite.Protocol/BgpConstants.cs:74`), but RFC 4486 §3 assigns:

- **4 — "Administrative Reset"**
- 6 — "Other Configuration Change"

**Wire evidence:** API peer deletion (`TerminatePeerAsync` → `NotifyCeaseAsync`) → the peer receives NOTIFICATION **6/6**, labeled "Other Configuration Change" instead of "Administrative Reset".

The neighboring constants are correct per the same table (1 = Maximum Number of Prefixes Reached, 2 = Administrative Shutdown, 3 = Peer De-configured) — only the Reset mapping is wrong, so every graceful reset / server shutdown / peer deletion mislabels its reason. Peers and NMS that key operator messaging off the subcode report the wrong cause.

## Fix shape
One constant: `CeaseAdministrativeReset = 4`. Grep for anything quoting the numeric value (docs/comments), update. Note: after this fix, `NotifyCeaseAsync` emits 6/4; the sniffing-stand scenario `delete-peer` asserts exactly that (reusable stand in `/tmp/bgpit/`, analyzer check F2).

## Severity
MEDIUM — correct teardown, wrong diagnostic label; every administrative reset on the wire is affected.

## Regression test
The existing session-level pattern: `NotifyCeaseAsync` on an Established scripted session → the sniffed/decoded NOTIFICATION is Cease **6/4**. (The integration analyzer check F2 already pins it; port a minimal variant into `BgpSessionShutdownTests`.)

## RFC
- RFC 4486 §3 (Cease subcode table — verified verbatim against the RFC text).

### LINKED ISSUE #507: tests: two CI timing flakes on loaded runners — RefreshDebounceTests (#302 signature recurrence) and PrefixAutoRefreshServiceTests, both rerun-green
## Found by
CI runs of the 2026-09-04 batch (#476–#488, tracking #489). Two distinct timing-test failures on loaded GitHub runners, both **green on rerun with zero changes**, both with diffs that cannot affect them (protocol-layer / providers-only). This is a two-test PATTERN now, not a one-off — hence a new issue rather than another "recurrence" comment (#302 lineage).

## Occurrences
| Where | Test | Failure | Rerun |
|---|---|---|---|
| PR #500 first run (build-test, job 100825100862) | `RefreshDebounceTests.ConcurrentRefreshes_CoalesceIntoOneCyclePlusOneLap` | `Assert.InRange (1–2), Actual: 3` — 15 ms | green, zero changes |
| PR #503 first run (build-test, job 100859833851) | `PrefixAutoRefreshServiceTests.PollsConditionalSourceMoreOftenThanNoEtagSource` | `Assert.Equal Expected 1, Actual 0` — 441 ms | green, zero changes |

Both were noted in PR comments at the time (#500, #503).

## Analysis
Both tests assert **cadence** — how many poll/debounce cycles happen within a wall-clock window — which on a loaded shared runner is scheduling-dependent: a slow tick starves the "more often" side (0 instead of 1) or lets a third refresh lap slip in (3 instead of ≤2). The `RefreshDebounceTests` one is a **recurrence of the #302 signature** (same assertion, same "Actual: 3"): it was fixed once and flaked again on #368 (2026-08-31, per the #302 comment trail), so the prior fix did not hold under runner load.

## Why existing protection is insufficient
The rest of the suite is FakeTimeProvider-driven or polls a condition (`SettleAsync`-style) instead of asserting counts per time window; these two assert counts, which is the fragile shape.

## Fix shape
- `PrefixAutoRefreshServiceTests.PollsConditionalSourceMoreOftenThanNoEtagSource`: drive the poller with `FakeTimeProvider` (the service already takes one) and assert on tick counts, not elapsed wall-clock; or poll the expected condition with a generous deadline.
- `RefreshDebounceTests.ConcurrentRefreshes_...`: the coalescing invariant ("one in-flight cycle + at most one pending lap") is what matters — assert it via completion-counting with a barrier/TaskCompletionSource control of cycle boundaries rather than a timing window; keep the timing window only as a generous outer bound.
- Optional CI hygiene: mark both `[Trait("Category", "TimingSensitive")]` so a future flake report is pre-classified.

## Severity
LOW-MEDIUM — test reliability only (false-red CI, manual reruns), no product defect implicated; two occurrences across two tests in one batch erodes CI trust.

## Regression test
The fixed tests themselves are the regression coverage (a flake repro = a red run under artificial load; determinism under FakeTimeProvider is the pass condition).

ruhex:
Three fixes found by the post-#503 local integration stand (sniffed traffic + RFC fact-checks) and the CI-flake pattern observed during the #476–#488 batch.

Closes #505
Closes #506
Closes #507

## Per-issue summary
- **#505** (PR #508, HIGH): `AwaitLoopTaskAsync` swallowed the cap-reset `BgpNotificationException`, so HoldTime>0 sessions sent a bare Cease **6/0** instead of **6/1** (RFC 4486 §2 "Maximum Number of Prefixes Reached") — wire-verified on the sniffing stand. Fix: one rethrow; red/green test (hold 30 session at cap → exactly one Cease with subcode 1).
- **#506** (PR #509, MEDIUM): `CeaseAdministrativeReset` was **6** ("Other Configuration Change") — RFC 4486 §3 assigns **4** ("Administrative Reset"); every graceful reset/shutdown/peer deletion was mislabeled on the wire. Fix: one constant; regression test asserts the LITERAL wire value 4.
- **#507** (PR #510): the two CI-flaky timing tests determinized — condition-waits replace the fixed 50 ms sleeps (the #503 CI failure shape), the jitter test interleaves clock advances with thread yields (scheduling-order immune), and the debounce test drains the initial route dump via a probe refresh before reading its baseline (the #500 "Actual: 3" shape — the dump's Load masqueraded as the refresh cycle's). 8/8 combined stability runs; assertions not weakened.

## Verification
- Per-PR: format/build 0w/0e, full suite green, dev CI green, CodeQL 0 alerts.
- Cumulative: 1095 tests; the stand's analyzer checks F1/F2 (sniffed 6/1 and 6/4) are the post-merge verification target.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected BGP Administrative Reset notifications to use the RFC-defined subcode.
  * Preserved max-prefix notification details when sessions exceed configured prefix limits, ensuring routes are properly withdrawn.

* **Tests**
  * Added coverage for administrative reset and max-prefix session shutdown behavior.
  * Improved refresh-related test reliability and determinism.

### DIFF
--- a/BGPLite.Protocol/BgpConstants.cs
+++ b/BGPLite.Protocol/BgpConstants.cs
@@ -71,7 +71,9 @@ public static class BgpConstants
         public const byte CeaseMaxPrefixes = 1;
         public const byte CeaseAdministrativeShutdown = 2;
         public const byte CeasePeerDeconfigured = 3;
-        public const byte CeaseAdministrativeReset = 6;
+        // #506: RFC 4486 §3 assigns 4 to "Administrative Reset" (6 is "Other Configuration
+        // Change") — the constant was 6, mislabeling every graceful reset on the wire.
+        public const byte CeaseAdministrativeReset = 4;
         public const byte CeaseConnectionRejected = 7;
     }
 
--- a/BGPLite.Server/BgpSession.cs
+++ b/BGPLite.Server/BgpSession.cs
@@ -756,6 +756,11 @@ public sealed class BgpSession : IDisposable
         try { await task; }
         catch (OperationCanceledException) { }
         catch (BgpParseException) { throw; }
+        // #505: a BgpNotificationException faulting a loop (the #304 cap reset thrown from
+        // HandleUpdateAsync on the read loop) must reach RunAsync's handler, which sends the
+        // carried code/subcode — Cease/MaxPrefixes per RFC 4486 §2. The generic catch below
+        // swallowed it, so HoldTime>0 sessions got a bare Cease 6/0 from the finally instead.
+        catch (BgpNotificationException) { throw; }
         catch (Exception ex) { _logger.LogWarning(ex, "{Label} loop faulted for {Peer}", label, _peer); }
     }
 
--- a/BGPLite.Tests/BgpSessionRouteOwnershipTests.cs
+++ b/BGPLite.Tests/BgpSessionRouteOwnershipTests.cs
@@ -743,6 +743,52 @@ public class BgpSessionRouteOwnershipTests
         Assert.Equal(0, routeTable.Count);                  // owned routes flushed by the finally
     }
 
+    /// <summary>
+    /// #505: the SAME cap-reset as <see cref="ExceedingMaxPrefixes_SendsCeaseMaxPrefixes_AndFlushesOwned"/>
+    /// but on a HoldTime &gt; 0 session — the read loop runs behind <c>Task.WhenAny</c>, and
+    /// <c>AwaitLoopTaskAsync</c>'s generic catch swallowed the cap's BgpNotificationException
+    /// (it only rethrew BgpParseException), so RunAsync's finally sent a bare Cease 6/0 and the
+    /// RFC 4486 §2 subcode never reached the peer. The rethrow restores the carried subcode here.
+    /// </summary>
+    [Fact]
+    public async Task ExceedingMaxPrefixes_WithHoldTimer_StillCarriesSubcode1()
+    {
+        var routeTable = new RouteTable();
+        var conn = new ScriptedConnection();
+        var config = new BgpConfig { Asn = 65001, RouterId = "127.0.0.1", HoldTime = 30, KeepAlive = 10, MaxPrefixesPerPeer = 2 };
+        var session = new BgpSession(
+            conn, new PeerConfig { Address = "127.0.0.1" }, config, routeTable,
+            AllowAllFilter.Instance, new BgpMetrics(), NullLogger<BgpSession>.Instance);
+        var run = session.RunAsync();
+        conn.EnqueueMessage(new BgpOpenMessage
+        {
+            Version = BgpConstants.BgpVersion,
+            Asn = 65002,
+            HoldTime = 30,
+            RouterId = 0x0A000002,
+            Capabilities = [BgpCapabilityInfo.FourOctetAsn(65002)],
+        });
+        conn.EnqueueMessage(BgpKeepaliveMessage.Instance);
+        for (var i = 0; i < 200 && !session.IsEstablished; i++)
+            await Task.Delay(TimeSpan.FromMilliseconds(10));
+        Assert.True(session.IsEstablished, "session must reach Established");
+
+        conn.EnqueueFrame(AnnounceFrame(8, 0x0A));                  // 1/2
+        conn.EnqueueFrame(AnnounceFrame(24, 0xC0, 0x00, 0x02));     // 2/2
+        conn.EnqueueFrame(AnnounceFrame(16, 0xC0, 0x01));           // 3rd distinct prefix — over
+        await run.WaitAsync(TimeSpan.FromSeconds(5));
+        Assert.False(session.IsEstablished);
+
+        var ceases = conn.Sent
+            .Select(b => BgpMessageReader.ReadMessage(b.AsSpan()))
+            .OfType<BgpNotificationMessage>()
+            .Where(n => n.ErrorCode == BgpConstants.Error.Cease)
+            .ToList();
+        var maxPrefixes = Assert.Single(ceases);                    // exactly one Cease on the wire
+        Assert.Equal(BgpConstants.SubError.CeaseMaxPrefixes, maxPrefixes.SubErrorCode);   // RED pre-fix: 0 (Unspecific)
+        Assert.Equal(0, routeTable.Count);                          // flush still happens
+    }
+
     /// <summary>#391: waits for the route table to reach the expected size. Generous timeout:
     /// a cold-JIT first run can take longer than SettleAsync's 2 s window to get from
     /// Established through the establish dump to a started read loop.</summary>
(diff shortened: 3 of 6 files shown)


ruhex:
Review dispositions (CodeRabbit on 87b95a2, 4 Minor): 3 fixed via #512 (now in this PR's head) — thread-safe counters, goal-driven advance-until-fired loop (tick period 600 s removes the ceiling); 1 dismissed — *wait for the initial dump before starting the probe* (RefreshDebounceTests): the ordering is already provable — the probe's RefreshCycleAsync can only acquire `_advertisedPrefixesLock` after the dump released it, so `await probe` completion strictly implies the dump (and its Load) finished before the baseline read; the `LoadCalls >= 1` early-wait is a progress hint only, not the guarantee. Re-review head: see checks.

ruhex:
CodeRabbit re-review of dc0d633 attempted, rate-limited (hourly quota). State: full review ran on 87b95a2 (covered head, Merge Risk: Low); the delta (dc0d633, via #512) contains exactly the 3 accepted fixes; the 4th finding dismissed with the lock-ordering proof above.

